PR #4, opened 2019-02-04
Code review for existing code

jathak:


lib/src/patch.dart
```diff
@@ -0,0 +1,57 @@
...
+    normalPatches.sort((a, b) => a.selection.compareTo(b.selection));
+    var buffer = StringBuffer();
+    int offset = 0;
+    patches
+        .where((p) => p.type == PatchType.prepend)
+        .map((p) => p.replacement)
+        .forEach(buffer.write);
+    for (var patch in normalPatches) {
+      if (patch.selection.start.offset < offset) {
+        throw new Exception("Can't apply overlapping patches.");
```
nex3:
```suggestion
        throw ArgumentError("Can't apply overlapping patches.");
```

If this ever actually gets thrown, it's going to be very painful to figure out why. Consider including [highlights](https://pub.dartlang.org/documentation/source_span/latest/source_span/SourceSpan/highlight.html) of the overlapping patches to indicate where they are.

lib/src/patch.dart
```diff
@@ -0,0 +1,57 @@
...
+    buffer.write(file.getText(offset));
+    patches
+        .where((p) => p.type == PatchType.append)
+        .map((p) => p.replacement)
+        .forEach(buffer.write);
+    return buffer.toString();
+  }
+}
+
+enum PatchType { normal, prepend, append }
```
nex3: I strongly recommend avoiding Dart's built-in enums. They don't support adding methods, overriding stringification, or implementing interfaces. [This pattern](https://github.com/sass/dart-sass/blob/7ca989a4f7dfa4afe269bf3a7aec887e9fa93900/lib/src/syntax.dart#L7-L36) is generally preferred.

lib/src/stylesheet_api.dart
```diff
@@ -0,0 +1,89 @@
...
+class Path {
+  final String path;
+  const Path(this.path);
+  toString() => path;
+  int get hashCode => path.hashCode;
+  operator ==(other) => path == other.path;
+}
+
+/// Takes a URL from an import rule and returns its canonical path.
+typedef PathResolver = Path Function(String importUrl, {Path from});
```
nex3: As of Dart 2, it's generally recommended to use inline function types rather than named function types, unless those types are very complex *and* likely to be used in many different places.

lib/src/stylesheet_api.dart
```diff
@@ -0,0 +1,89 @@
...
+
+/// Returns the contents of the file at [path].
+typedef FileLoader = String Function(Path path);
+
+// The API of a stylesheet, which consists of its variables, functions, mixins,
+// and imported dependencies.
+class StylesheetApi {
+  final Stylesheet sheet;
```
nex3: I think it's a little confusing to refer to a stylesheet as just a "sheet"; to my knowledge, no other CSS-related technology does that.

lib/src/stylesheet_api.dart
```diff
@@ -0,0 +1,89 @@
...
+  Map<String, FunctionRule> functions = {};
+
+  Map<String, MixinRule> mixins = {};
+
+  Map<Path, StylesheetApi> imports = {};
+
+  StylesheetApi._(this.sheet);
+
+  factory StylesheetApi(Path path, PathResolver resolver, FileLoader loader,
+      {Map<Path, StylesheetApi> existingApis}) {
```
nex3: Oh, now I get why you're constructing a bunch of these. Having a constructor destructively modify one of its arguments is *very* squicky to me... I strongly recommend refactoring to avoid this. For example, if you keep track of all global members and the paths where they're defined, you may not need this at all.

test/migrations/README.md
```diff
@@ -0,0 +1,31 @@
+# Migration Tests
+
+Each set of source files used for a test of the migrator is represented a single
+[HRX archive](https://github.com/google/hrx).
+
+> Note: The test script does not currently use a proper HRX parser, so `<==>` is
+> only boundary allowed.
```
nex3:
```suggestion
> the only boundary allowed.
```

test/migrations/functions.hrx
```diff
@@ -0,0 +1,33 @@
...
+<==> expected/_functions.scss
+@use "sass:list";
+$twelve: 12;
+
+@function plus($a, $b) {
+  @return $a + $b;
+}
+@function subtract_all($first, $rest...) {
+  @return $first - list.nth($rest, 1) - list.nth($rest, 2) - list.nth($rest, 3);
+}
```
nex3: I recommend as much as possible trying to make tightly-scoped test cases that exercise exactly one piece of functionality. That'll make it easier to tell at a glance what functionality is tested, to ensure that test failures don't hide one another, and to ensure that modifying a test case doesn't accidentally remove something it was trying to test.

test/migrations/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+Each HRX archive should have two root directories, `input` and `expected`, and
+optionally one file at the root, `recursive_manifest`.
+
+All source files to be migrated should live in `input` (or its subdirectories).
+These files should not attempt to import any other files outside of the `input`
+directory of this HRX archive.
+
+For each source file in `input`, there should be a corresponding file in
+`expected` that represents the expected output of the migrator for that file.
```
nex3: Consider "output" instead of "expected", for symmetry with "input".

test/migrations/variables.hrx
```diff
@@ -0,0 +1,63 @@
...
+
+<==> expected/three.scss
+$b: blue;
+$c: 5px !default;
+b {
+  width: 0;
+}
+
+<==> expected/four.scss
+$c: 4px;
```
nex3: At some point, it would be good to remove variable declarations that are used to configure an import. Maybe file an issue for that?

test/migrator_test.dart
```diff
@@ -0,0 +1,114 @@
...
+        filename = line.substring(5).trim();
+        contents = "";
+      } else {
+        contents += line + "\n";
+      }
+    }
+    if (filename != null) _load(filename, contents);
+  }
+
+  _load(String filename, String contents) {
```
nex3:
```suggestion
  void _load(String filename, String contents) {
```

Some legacy code uses an untyped return to mean "can return any type", so it's good to be explicit when you mean "doesn't return anything".

test/migrator_test.dart
```diff
@@ -0,0 +1,114 @@
...
+  Map<String, String> testFiles = {};
+  Map<String, String> expectedOutput = {};
+  Map<String, List<String>> recursiveManifest = {};
+
+  HrxTestFiles(String hrxName) {
+    var hrxText = File("test/migrations/$hrxName.hrx").readAsStringSync();
+    // TODO(jathak): Replace this with an actual HRX parser.
+    String filename;
+    String contents;
+    for (String line in hrxText.substring(0, hrxText.length - 1).split("\n")) {
```
nex3:
```suggestion
    for (var line in hrxText.substring(0, hrxText.length - 1).split("\n")) {
```

[Avoid type annotating initialized local variables](https://www.dartlang.org/guides/language/effective-dart/design#avoid-type-annotating-initialized-local-variables).

test/patch_test.dart
```diff
@@ -0,0 +1,41 @@
...
+import 'package:sass_module_migrator/src/patch.dart';
+import 'package:source_span/source_span.dart';
+import 'package:test/test.dart';
+
+void main() {
+  test("single patch applied", () {
+    var file = SourceFile.fromString("abcde");
+    var patch = Patch(file.span(2, 4), "fgh");
+    expect(Patch.applyAll(file, [patch]), equals("abfghe"));
+  });
```
nex3: Nit: if you add empty lines between these tests, it'll help the file feel less cramped.

test/patch_test.dart
```diff
@@ -0,0 +1,41 @@
...
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import 'package:sass_module_migrator/src/patch.dart';
+import 'package:source_span/source_span.dart';
+import 'package:test/test.dart';
+
+void main() {
+  test("single patch applied", () {
```
nex3: Generally speaking, test names should be full sentences that succinctly describe the expected behavior. For example, "a single patch replaces the original text".

jathak:
Thank you for such a thorough review! I'll probably work through these in stages as I have time.

lib/src/migrator.dart
```diff
@@ -0,0 +1,296 @@
...
+  /// Adds a namespace to an include rule if it is necessary.
+  @override
+  void visitIncludeRule(IncludeRule node) {
+    var ns = findNamespaceFor(node.name, ApiType.mixins);
+    if (ns == null) {
+      ns = makeImplicitDependencyExplicit(node.name, ApiType.mixins);
+      if (ns == null) return;
+    }
+    var endName = node.arguments.span.start.offset;
+    var startName = endName - node.name.length;
```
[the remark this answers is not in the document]
jathak: `node.span.start.offset` would start at `@include` instead of at the beginning of the name. I wanted to start at the name instead to only patch the name and allow the shorthand syntax for `.sass` to be preserved if it was there in the original stylesheet.

jathak:
I refactored away `BaseVisitor` and `StylesheetApi`, changing the migrator itself to run a single pass on all dependencies, with each stylesheet's migration state represented by a new `StylesheetMigration` class. This also allowed the scoping and implicit import tracking to be simplified in my mind.

I believe I've addressed most of your comments, aside from more tightly-scoped tests, which I'll work on in a separate branch. Let me know if anything needs clarification. Thanks!

lib/src/migrator.dart
```diff
@@ -0,0 +1,247 @@
...
+    return results;
+  }
+
+  /// Migrates the stylesheet at [path] if it hasn't already been migrated and
+  /// returns the StylesheetMigration instance for it regardless.
+  StylesheetMigration _migrateStylesheet(String path) {
+    path = canonicalizePath(path,
+        loadPath: _currentMigration == null
+            ? p.current
+            : p.dirname(_currentMigration.path));
```
nex3: Setting the load path to the directory of the current file is probably right. I'm guessing you're doing that to resolve relative imports, but they should instead be handled by interpreting `path` as relative to the entrypoint's path. So, you'd call:

```dart
path = canonicalizePath(
    _currentMigration == null ? path : p.join(p.dirname(_currentMigration.path), path));
```

Note that `p.join()` is a no-op if the final argument is already absolute.

You shouldn't need to set the load path at all in here; or rather, you should probably eventually have a `--load-path` CLI argument to match the executable's, but until then you can just assume there's no load path and all imports are relative.

lib/src/migrator.dart
```diff
@@ -0,0 +1,247 @@
...
+    if (_migrations.containsKey(path)) return _migrations[path];
```
nex3: I think you can clean this up a bit using [`Map.putIfAbsent()`](https://api.dartlang.org/stable/2.1.0/dart-core/Map/putIfAbsent.html).

lib/src/migrator.dart
```diff
@@ -0,0 +1,247 @@
...
+    _activeMigrations.remove(migration);
+    return migration;
+  }
+
+  /// Visits the children of [node] with a local scope.
+  ///
+  /// Note: The children of a stylesheet are at the root, so we should not add
+  /// a local scope.
+  @override
+  visitChildren(ParentStatement node) {
```
nex3:
```suggestion
  void visitChildren(ParentStatement node) {
```

lib/src/migrator.dart
```diff
@@ -0,0 +1,247 @@
...
+    if (namespace == null) {
+      if (!builtInFunctionModules.containsKey(name)) return;
+      namespace = builtInFunctionModules[name];
+      name = builtInFunctionNameChanges[name] ?? name;
+      if (!_currentMigration.additionalUseRules.contains("sass:$namespace")) {
+        _currentMigration.additionalUseRules.add("sass:$namespace");
+      }
+    }
+    _currentMigration.patches.add(Patch(node.name.span, "$namespace.$name"));
+  }
```
nex3: A few judiciously-placed newlines would probably make it easier to understand the flow of this function.

lib/src/migrator.dart
```diff
@@ -0,0 +1,247 @@
...
+    visitInterpolation(node.name);
+    visitArgumentInvocation(node.arguments);
+    if (node.name.asPlain == null) return;
+    var name = node.name.asPlain;
+    if (_currentMigration.localScope?.isLocalFunction(name) ?? false) return;
+    var namespace;
+    if (_currentMigration.functions.containsKey(name)) {
+      namespace =
+          _currentMigration.namespaceForNode(_currentMigration.functions[name]);
+    }
```
nex3: I'd write:

```dart
var namespace = _currentMigration.functions.containsKey(name)
    ? _currentMigration.namespaceForNode(_currentMigration.functions[name])
    : null;
```

This isn't just style; as currently written, Dart's type inference (which doesn't look outside individual statements) is unable to determine the type of `namespace`, so it ends up as `dynamic`, which could make analysis worse down the line.

lib/src/migrator.dart
```diff
@@ -0,0 +1,247 @@
...
+    if (namespace == null) {
+      if (!builtInFunctionModules.containsKey(name)) return;
```
nex3: I wonder if it might be a good idea to emit a warning here. If the migrator can't tell what a function is referring to, chances are the user invoked it wrong somehow and wants to know about it. Not something you need to do right now, but probably a good idea to add TODOs so you can easily find the places to add warnings down the line.

lib/src/migrator.dart
```diff
@@ -0,0 +1,247 @@
...
+    var config = "";
+    if (overrides.isNotEmpty) {
+      config = " with (\n  " + overrides.join(',\n  ') + "\n)";
+    }
+    _currentMigration.patches
+        .add(Patch(node.span, '@use ${import.span.text}$config'));
+    _currentMigration.namespaces[importMigration.path] =
+        namespaceForPath(import.url);
+
+    // Ensure that references to members transient dependencies can be
```
nex3:
```suggestion
    // Ensure that references to members' transient dependencies can be
```

lib/src/migrator.dart
```diff
@@ -0,0 +1,247 @@
...
+    // namespaced.
+    _currentMigration.variables.addEntries(importMigration.variables.entries);
+    _currentMigration.mixins.addEntries(importMigration.mixins.entries);
+    _currentMigration.functions.addEntries(importMigration.functions.entries);
```
nex3: Are these necessary to track now that you're also tracking the global scope? In fact, won't they be misleading, since once the current stylesheet is migrated to `@use` it *won't* make these members transitively visible?

lib/src/migrator.dart
```diff
@@ -0,0 +1,247 @@
...
+    if (node is Stylesheet) {
+      super.visitChildren(node);
+      return;
+    }
+    _currentMigration.localScope = LocalScope(_currentMigration.localScope);
+    super.visitChildren(node);
+    _currentMigration.localScope = _currentMigration.localScope.parent;
+  }
+
+  /// Adds a namespace to any function calls that require them.
```
nex3:
```suggestion
  /// Adds a namespace to any function call that requires it.
```

(Also below)

lib/src/stylesheet_migration.dart
```diff
@@ -0,0 +1,132 @@
...
+      functions[node.name] = node;
+    } else {
+      localScope.functions.add(node.name);
+    }
+  }
+
+  /// Finds the namespace for the stylesheet containing [node], adding a new use
+  /// rule if necessary.
+  String namespaceForNode(SassNode node) {
+    var nodePath = node.span.sourceUrl.path;
```
nex3:
```suggestion
    var nodePath = p.fromUri(node.span.sourceUrl);
```

A `file:` URL's path component isn't always identical to a filesystem path.

lib/src/migrator.dart
jathak: `_currentMigration.functions` only contains functions defined in the global scope (whether in the current file or in one of its imports), so it's possible this could be referring to a local function.

lib/src/migrator.dart
jathak: By adding the global members from imports to the current stylesheet, it lets us find existing references to transitively-visible members that need to be namespaced with the transitive dependency, not the direct one. This allows `namespaceForNode` to return the correct namespace (adding it if it hasn't been `@use`-d) by looking at the source path of the `VariableDeclaration`, `MixinRule`, or `FunctionRule` in one of these dictionaries.

jathak:
Addressed remaining comments. See my explanation of why we need to copy the global members of an import into the current stylesheet's member maps [here](https://github.com/sass/module_migrator/pull/4#discussion_r256610417)

lib/src/migrator.dart
nex3: Oh, because the scope is bound to a `StylesheetMigration`? Maybe it should be a member of `_Migrator` instead... after all, the same global scope is shared across all stylesheets.

lib/src/migrator.dart
```diff
@@ -0,0 +1,248 @@
...
+import 'utils.dart';
+
+/// Runs a migration of [entrypoints] and their dependencies without writing
+/// any changes to disk.
+p.PathMap<String> migrateFiles(List<String> entrypoints) =>
+    _Migrator().migrate(entrypoints);
+
+class _Migrator extends RecursiveStatementVisitor implements ExpressionVisitor {
+  /// List of all migrations for files touched by this run.
+  final _migrations = p.PathMap();
```
nex3:
```suggestion
  final _migrations = p.PathMap<StylesheetMigration>();
```

lib/src/migrator.dart
```diff
@@ -0,0 +1,248 @@
...
+
+  /// List of migrations in progress. The last item is the current migration.
+  List<StylesheetMigration> _activeMigrations = [];
```
nex3:
```suggestion
  final _activeMigrations = <StylesheetMigration>[];
```

lib/src/migrator.dart
```diff
@@ -0,0 +1,248 @@
...
+
+  /// Migrates the stylesheet at [path] if it hasn't already been migrated and
+  /// returns the StylesheetMigration instance for it regardless.
+  StylesheetMigration _migrateStylesheet(String path) {
+    path = canonicalizePath(_currentMigration == null
+        ? path
+        : p.join(p.dirname(_currentMigration.path), path));
+    return _migrations.putIfAbsent(path, () {
+      var migration = StylesheetMigration(path);
+      _migrations[path] = migration;
```
nex3: This is no longer necessary; `putIfAbsent` takes care of adding the migration once the block is complete.

lib/src/migrator.dart
jathak: I don't think a single global scope would work when the migrator is running on multiple entrypoints. The scope would need to be cleared between each entrypoint, but the stored global members for each dependency need to remain available in case a future entrypoint also re-uses them.

nex3: I suspect that you'll end up needing to abandon shared state across entrypoints. The way names are resolved in an individual file is highly dependent on the global state at the time it's loaded, so I suspect you'll run into lots of nasty edge cases if you don't treat each entrypoint as an entirely separate run.

jathak: Oh, okay. That makes sense. I'll change it so the scoping information is just stored in `_Migrator`.

What should I do in the case where the user wants to migrate two entrypoints that share a dependency, but the dependency would be migrated differently depending on which entrypoint is loading it?

nex3: Probably just throw an error.

jathak:
I changed the tests to be more tightly scoped and addressed your remaining comments. Let me know if there's anything else I should do before merging #8.

PR #13, opened 2019-03-09
Basic support for namespacing get-function calls

### LINKED ISSUE #11: Add namespaces to `get-function()` calls
For those whose referent can't be statically determined, print warnings.

jathak:
Resolves #11.

If the argument passed to get-function is a string literal, we'll add a namespace if necessary.

For all other `get-function` calls, we'll emit a warning saying that a namespace may be required. We might eventually be able to detect some of these with some more sophisticated logic here, but for now we'll just handle the basics.

### DIFF
--- a/lib/src/migrator.dart
+++ b/lib/src/migrator.dart
@@ -127,35 +127,44 @@ class _Migrator extends RecursiveStatementVisitor implements ExpressionVisitor {
   /// Adds a namespace to any function call that require it.
   void visitFunctionExpression(FunctionExpression node) {
     visitInterpolation(node.name);
-    _visitFunctionName(node.name.asPlain, node.name.span);
+    _patchNamespaceForFunction(node.name.asPlain, (name, namespace) {
+      _currentMigration.patches.add(Patch(node.name.span, "$namespace.$name"));
+    });
     visitArgumentInvocation(node.arguments);
 
     if (node.name.asPlain == "get-function") {
-      var fnName = node.arguments.named['name'];
-      if (fnName == null) {
-        if (node.arguments.positional.isEmpty) {
-          throw ArgumentError("No name passed to $node");
-        } else {
-          fnName = node.arguments.positional.first;
-        }
+      var nameArgument =
+          node.arguments.named['name'] ?? node.arguments.positional.first;
+      if (nameArgument is! StringExpression ||
+          (nameArgument as StringExpression).text.asPlain == null) {
+        print(nameArgument.span.message(
+            "WARNING - get-function call may require \$module parameter"));
+        return;
       }
-      if (fnName is StringExpression && fnName.text.asPlain != null) {
+      var fnName = nameArgument as StringExpression;
+      _patchNamespaceForFunction(fnName.text.asPlain, (name, namespace) {
         var span = fnName.span;
         if (fnName.hasQuotes) {
           span = span.file.span(span.start.offset + 1, span.end.offset - 1);
         }
-        _visitFunctionName(fnName.text.asPlain, span);
-      } else {
-        _warn(
-            "WARNING: Could not determine if namespace is required for "
-            "get-function call @",
-            fnName.span);
-      }
+        _currentMigration.patches.add(Patch(span, name));
+        var beforeParen = node.span.end.offset - 1;
+        _currentMigration.patches.add(Patch(
+            node.span.file.span(beforeParen, beforeParen),
+            ", \$module: $namespace"));
+      });
     }
   }
 
-  /// Adds a namespace to function [name] if it requires one.
-  void _visitFunctionName(String name, FileSpan span) {
+  /// Calls [patcher] when the function [name] requires a namespace and adds a
+  /// new use rule if necessary.
+  ///
+  /// [patcher] takes two arguments: the name used to refer to that function
+  /// when namespaced, and the namespace itself. The name will match the name
+  /// provided to the outer function except for built-in functions whose name
+  /// within a module differs from its original name.
+  void _patchNamespaceForFunction(
+      String name, void patcher(String name, String namespace)) {
     if (name == null) return;
     if (_localScope?.isLocalFunction(name) ?? false) return;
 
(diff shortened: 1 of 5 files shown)


lib/src/migrator.dart
```diff
@@ -126,10 +127,36 @@ class _Migrator extends RecursiveStatementVisitor implements ExpressionVisitor {
...
+      if (fnName is StringExpression && fnName.text.asPlain != null) {
+        var span = fnName.span;
+        if (fnName.hasQuotes) {
+          span = span.file.span(span.start.offset + 1, span.end.offset - 1);
+        }
+        _visitFunctionName(fnName.text.asPlain, span);
+      } else {
+        _warn(
+            "WARNING: Could not determine if namespace is required for "
+            "get-function call @",
```
nex3: Nit: What's this `@` for?

lib/src/migrator.dart
```diff
@@ -143,7 +170,7 @@ class _Migrator extends RecursiveStatementVisitor implements ExpressionVisitor {
       name = builtInFunctionNameChanges[name] ?? name;
       _currentMigration.additionalUseRules.add("sass:$namespace");
     }
-    _currentMigration.patches.add(Patch(node.name.span, "$namespace.$name"));
+    _currentMigration.patches.add(Patch(span, "$namespace.$name"));
```
nex3: It doesn't actually say this for `get-function()` in the spec yet (oops), but the way [other functions work](https://github.com/sass/language/blob/master/accepted/module-system.md#new-features-for-existing-functions) and the way `get-function()` *should* work is taking an additional optional `$module` parameter that specifies the module, rather than parsing the module name out of the function name.

lib/src/migrator.dart
```diff
@@ -276,6 +303,13 @@ class _Migrator extends RecursiveStatementVisitor implements ExpressionVisitor {
     }
   }
 
+  /// Emits a warning [message] with [span] and its context.
+  void _warn(String message, FileSpan span) {
+    print(message);
+    print(span.highlight());
```
nex3: You probably want to use [`SourceSpan.message`](https://pub.dartlang.org/documentation/source_span/latest/source_span/SourceSpan/message.html) here; it's built to associate messages with spans.

test/migrations/namespace_get_function.hrx
```diff
@@ -0,0 +1,39 @@
...
+<==> input/_library.scss
+@function increment($x) {
+  @return $x + 1;
+}
+
+<==> output/entrypoint.scss
+@use "sass:meta";
+@use "library";
+
+$fn1: meta.get-function(library.increment);
```
nex3: I don't think it matters given the new module syntax, but FYI `library.increment` isn't a valid SassScript expression. `.` isn't a valid identifier character.

lib/src/migrator.dart
```diff
@@ -126,10 +127,45 @@ class _Migrator extends RecursiveStatementVisitor implements ExpressionVisitor {
...
+      _patchNamespaceForFunction(fnName.text.asPlain, (name, namespace) {
+        var span = fnName.span;
+        if (fnName.hasQuotes) {
+          span = span.file.span(span.start.offset + 1, span.end.offset - 1);
+        }
+        _currentMigration.patches.add(Patch(span, name));
+        var beforeParen = node.span.end.offset - 1;
+        _currentMigration.patches.add(Patch(
+            node.span.file.span(beforeParen, beforeParen),
+            ", \$module: $namespace"));
```
nex3: Nit: Emit quotes around the namespace, in case it's a color name or something. There is actually [a Sass library named `true`](https://github.com/oddbird/true), so this isn't even purely hypothetical.

test/migrations/namespace_get_function.hrx
```diff
@@ -0,0 +1,37 @@
...
+line 5, column 20 of $TEST_DIR/entrypoint.scss: WARNING - get-function call may require $module parameter
+  ,
+5 | $fn3: get-function("incre" + "ment");
+  |                    ^^^^^^^^^^^^^^^^
+  '
+line 8, column 20 of $TEST_DIR/entrypoint.scss: WARNING - get-function call may require $module parameter
+  ,
+8 | $fn4: get-function($x);
+  |                    ^^
+  '
```
nex3: Also test a getting a function from a core library when the function's namespaced name doesn't match its original name, and getting a function from a library named `true` or whatever.

lib/src/migrator.dart
```diff
@@ -126,10 +127,45 @@ class _Migrator extends RecursiveStatementVisitor implements ExpressionVisitor {
...
 
-    if (node.name.asPlain == null) return;
-    var name = node.name.asPlain;
+    if (node.name.asPlain == "get-function") {
+      var nameArgument =
+          node.arguments.named['name'] ?? node.arguments.positional.first;
+      if (nameArgument is! StringExpression ||
+          (nameArgument as StringExpression).text.asPlain == null) {
+        print(nameArgument.span.message(
+            "WARNING - get-function call may require \$module parameter"));
```
nex3: It's probably a good idea to print this to stderr.

test/migrations/namespace_get_function.hrx
```diff
@@ -0,0 +1,47 @@
+<==> input/entrypoint.scss
+@import "library";
+@import "true";
```
nex3: Add a comment explaining why this is an interesting case.

lib/src/migrator.dart
jathak: Is there an equivalent to the `prints` matcher for testing stderr?

nex3: Unfortunately no there isn't. Usually when I'm testing command-line executables, I just run them via `Process.runSync()` or [the `test_process` package](https://pub.dartlang.org/packages/test_process) so I don't have to worry about mocking out all the APIs.

jathak: Okay, in that case, I'll leave this as is for now. I've made an issue (#14) for this, which I'll handle in a separate PR

PR #16, opened 2019-04-01
Generalizes to support multiple migrators

### LINKED ISSUE #19: Generalize tool to support multiple migrators

jathak:
Resolves #19.

There's now a `Migrator` class that additional migrators can extend.

Most migrators will want to extend `MigratorBase`, which handles the tree traversal. For now, `MigratorBase` only operates on a single stylesheet, with the dependency migration implemented only by `ModuleMigrator`.


lib/src/migrator_base.dart
```diff
@@ -0,0 +1,113 @@
...
+
+/// This base migrator parses the stylesheet at each provided entrypoint and
+/// recursively visits every statement and expression it contains.
+///
+/// Migrators based on this should add appropriate patches to [patches] in
+/// overridden methods.
+///
+/// On its own, this migrator only touches the entrypoints that are directly
+/// provided to it; it does not migrate dependencies.
+abstract class MigratorBase extends RecursiveStatementVisitor
```
nex3: Calling this `MigratorBase` doesn't communicate what the difference is between it and `Migrator`. Maybe `AstMigrator` to indicate that it traverses the AST?

lib/src/migrator_base.dart
```diff
@@ -0,0 +1,113 @@
...
+    with Migrator, ExpressionVisitor {
+  List<Patch> _patches = [];
```
nex3: Nit: When I have a getter that exposes a field to make it read-only, I usually write the field directly underneath the getter to make it obvious that they're associated. It's also a good idea to wrap the getter in an `UnmodifiableListView` to ensure that external users don't accidentally call destructive methods on it. For example:

```dart
  /// The patches to be applied to the stylesheet being migrated.
  ///
  /// Subclasses that override this should also override [migrateFile].
  List<Patch> get patches => UnmodifiableListView(_patches);
  List<Patch> _patches = [];
```

lib/src/migrator_base.dart
```diff
@@ -0,0 +1,113 @@
...
+    if (_patches.isNotEmpty) {
+      results[path] = Patch.applyAll(_patches.first.selection.file, _patches);
+    }
+    _patches = null;
+    return results;
+  }
+
+  // Expression Tree Traversal
+
+  @override
```
nex3: You seem to be using this inconsistently... all of the functions below are overridden, but only this is tagged. Either use it for all overrides or none.

lib/src/migrator_base.dart
```diff
@@ -0,0 +1,113 @@
...
+  visitExpression(Expression expression) => expression.accept(this);
+
+  visitBinaryOperationExpression(BinaryOperationExpression node) {
+    node.left.accept(this);
+    node.right.accept(this);
```
nex3: What do you think about adding an abstract class to the `sass` package that implements both `RecursiveStatementVisitor` and `ExpressionVisitor`? It seems silly to have to implement this very basic logic of "visit every node in the AST" multiple times.

lib/src/migrators/module.dart
```diff
@@ -0,0 +1,302 @@
+// Copyright 2018 Google LLC
```
```suggestion
// Copyright 2019 Google LLC
```

lib/src/migrators/module.dart
```diff
@@ -0,0 +1,302 @@
...
+
+import 'package:sass_migrator/src/migrator_base.dart';
+import 'package:sass_migrator/src/patch.dart';
+import 'package:sass_migrator/src/utils.dart';
+
+import 'module/built_in_functions.dart';
+import 'module/local_scope.dart';
+import 'module/stylesheet_migration.dart';
+
+class ModuleMigrator extends MigratorBase {
```
nex3: Document this.

lib/src/migrators/module.dart
```diff
@@ -0,0 +1,302 @@
...
+  final name = "module";
+  final description = "Migrates stylesheets to the new module system.";
+  final argParser = ArgParser()
+    ..addFlag('migrate-deps',
+        abbr: 'd', help: 'Migrate dependencies in addition to entrypoints.');
+
+  bool get _migrateDependencies => argResults['migrate-deps'];
```
nex3: It's probably a good idea to explicitly add `as bool` here. Otherwise, you're relying on an implicit cast, which is likely to become an analysis error at some point.

Also, it's up to you, but I generally prefer to just refer to arguments directly rather than creating indirection through a getter.

test/migrator_test.dart
```diff
@@ -16,26 +17,34 @@ import 'package:test_descriptor/test_descriptor.dart' as d;
 /// Runs all migration tests. See migrations/README.md for details.
 void main() {
   glyph.ascii = true;
-  var migrationTests = Directory("test/migrations");
-  for (var file in migrationTests.listSync().whereType<File>()) {
-    if (file.path.endsWith(".hrx")) {
-      test(p.basenameWithoutExtension(file.path), () => testHrx(file));
+  testMigrator(ModuleMigrator());
```
nex3: I recommend creating a separate entrypoint file for each migrator you want to test, rather than putting everything in a single file. That'll make it a lot easier to just test one migrator at a time.

lib/src/migrator_base.dart
jathak: Yeah, that makes sense. Created sass/dart-sass#639 to add this.

lib/src/migrator.dart
```diff
@@ -1,362 +1,56 @@
...
+  /// included in the results.
+  p.PathMap<String> migrateFiles(Iterable<String> entrypoints) {
+    var allMigrated = p.PathMap<String>();
+    for (var entrypoint in entrypoints) {
+      var migrated = migrateFile(entrypoint);
+      for (var file in migrated.keys) {
+        if (allMigrated.containsKey(file) &&
+            migrated[file] != allMigrated[file]) {
+          throw UnsupportedError(
+              "$file is migrated in more than one way by these entrypoints.");
```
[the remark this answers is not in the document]
jathak: Created #23 and #24 for this.

jathak:
I've updated this to now have the generic migration support split into `Migrator`, which extends `Command` and contains state for the entire migration run, and `MigrationVisitor`, which migrators subclass and use to traverse the AST of a single stylesheet.

I've also made the `--migrate-deps` flag global. The module migrator has some special logic here since it needs to visit all dependencies even if they're not being migrated, but any migrators that don't rely on cross-stylesheet information should be able to support migrating dependencies without any additional code.

lib/src/migrator.dart
```diff
@@ -4,48 +4,62 @@
...
+/// Subclasses need to implement [migrateFile], which takes an entrypoint,
+/// migrates it, and stores the results in [migrated]. If [migrateDependencies]
+/// is true, they should also migrate all of that entrypoint's direct and
+/// indirect dependencies.
+///
+/// Most migrators will want to create a subclass of [MigrationVisitor] and
+/// implement [migrateFile] with `MyMigrationVisitor(this, entrypoint).run()`.
+abstract class Migrator extends Command<p.PathMap<String>> {
+  /// The entrypoints that this migrator will run from.
+  List<String> get entrypoints => argResults.rest;
```
nex3: Why is this a (public) getter? Are you expecting a subclass to override it?

lib/src/migrator.dart
```diff
@@ -4,48 +4,62 @@
...
 
-  /// Parser for the arguments that this migrator takes.
-  ArgParser get argParser => ArgParser();
+  /// If true, dependencies will be migrated in addition to the entrypoints.
+  bool get migrateDependencies => globalResults['migrate-deps'] as bool;
 
-  /// Set by the executable based on [argParser] and the provided arguments.
-  ArgResults argResults;
+  /// Migrated contents of stylesheets that have already been migrated.
+  final migrated = p.PathMap<String>();
```
nex3: Once `MigrationVisitor` no longer modifies this, I think you can do away with it as a field and just make it the return value of `run()`.

lib/src/migrator.dart
```diff
@@ -4,48 +4,62 @@
...
-  /// migrator and its configuration.
+  /// Runs this migrator on [entrypoint] (and its dependencies, if the
+  /// --migrate-deps flag is passed).
   ///
-  /// Files that did not require migration, even if touched by the migrator,
+  /// Files that did not require any changes, even if touched by the migrator,
   /// should not be included map of results.
-  p.PathMap<String> migrateFile(String entrypoint);
+  @protected
+  void migrateFile(String entrypoint);
```
nex3: As with `MigrationVisitor.run()`, I think it's cleaner if this returns a value rather than modifying a field.

lib/src/migrator.dart
```diff
@@ -4,48 +4,62 @@
...
   ///
   /// Entrypoints and dependencies that did not require any changes will not be
   /// included in the results.
-  p.PathMap<String> migrateFiles(Iterable<String> entrypoints) {
+  p.PathMap<String> run() {
     var allMigrated = p.PathMap<String>();
     for (var entrypoint in entrypoints) {
-      var migrated = migrateFile(entrypoint);
+      migrated.clear();
+      migrateFile(canonicalizePath(p.join(Directory.current.path, entrypoint)));
```
nex3: If you're doing `p.join(Directory.current.path, ...)`, something smells wrong. Anything that handles paths should transparently treat a relative path as relative to the current directory. Hopefully moving to using `FilesystemImporter` makes this unnecessary too?

lib/src/migrators/module.dart
```diff
@@ -1,41 +1,30 @@
...
 
 import 'module/built_in_functions.dart';
 import 'module/local_scope.dart';
-import 'module/stylesheet_migration.dart';
 
-class ModuleMigrator extends MigratorBase {
+/// Migrates stylesheets to the new module system.
+class ModuleMigrator extends Migrator {
   final name = "module";
+  final aliases = ["modules", "module-system"];
```
nex3: I'm personally not a fan of adding command aliases unless you have to for backwards compatibility (or puns, like `pub lish` :stuck_out_tongue_winking_eye:), but it's up to you.

lib/src/migrators/module.dart
```diff
@@ -46,54 +35,55 @@ class ModuleMigrator extends MigratorBase {
...
+  LocalScope localScope;
+
+  /// Returns the migrated contents of this stylesheet, based on [patches] and
+  /// [additionalUseRules], or null if the stylesheet does not change.
+  @override
+  String getMigratedContents() {
+    if (patches.isEmpty && additionalUseRules.isEmpty) return null;
+    var semicolon = syntax == Syntax.sass ? "" : ";";
+    var uses = additionalUseRules.map((use) => '@use "$use"$semicolon\n');
+    var contents = Patch.applyAll(stylesheet.span.file, patches);
```
nex3: Is there a reason to avoid `super.getMigratedContents()` here?

lib/src/migrators/module.dart
```diff
@@ -46,54 +35,55 @@ class ModuleMigrator extends MigratorBase {
...
+  final ModuleMigrator migrator;
+  final String path;
+
+  _ModuleMigrationVisitor(this.migrator, this.path);
+
+  _ModuleMigrationVisitor newInstance(String newPath) =>
+      _ModuleMigrationVisitor(migrator, newPath);
+
+  /// Namespaces of modules used in this stylesheet.
+  final namespaces = p.PathMap<String>();
```
nex3: I'd keep these members private. Even though the class itself is private now, making its members private as well helps communicate that they aren't intended to be used outside this class (technically they *can* be, but readers use a private member as a mental shortcut that indicates class privacy).

lib/src/utils.dart
```diff
@@ -43,3 +43,18 @@ Patch patchAfter(AstNode node, String text) {
 void warn(String message, FileSpan context) {
   print(context.message("WARNING - $message"));
 }
+
+class MigrationException {
```
nex3: Document this.

lib/src/utils.dart
```diff
@@ -43,3 +43,18 @@ Patch patchAfter(AstNode node, String text) {
...
+  final String message;
+  final FileSpan context;
```
nex3: In most places, we just call the span associated with a class `span`.

Also, document this.

lib/src/utils.dart
```diff
@@ -22,3 +26,35 @@ String namespaceForPath(String path) {
...
+  print(context.message("WARNING - $message"));
+}
+
+class MigrationException {
+  final String message;
+  final FileSpan context;
+
+  MigrationException(this.message, {this.context});
+
+  toString() {
```
```suggestion
  String toString() {
```

pubspec.yaml
```diff
@@ -6,8 +6,10 @@ environment:
   sdk: '>=2.1.0 <3.0.0'
 
 dependencies:
-  args: "^1.4.0"
-  sass: "^1.17.3"
+  args: "^1.5.1"
+  sass:
+    # Temporary, until 1.17.5 is on pub
+    git: git://github.com/sass/dart-sass
```
nex3: Use [dependency overrides](https://www.dartlang.org/tools/pub/dependencies#dependency-overrides) for temporary dependencies.

lib/src/migration_visitor.dart
```diff
@@ -0,0 +1,101 @@
...
+
+  /// Runs the migrator and stores the migrated contents in `migrator.migrated`.
+  void run() {
+    var contents = File(path).readAsStringSync();
+    syntax = Syntax.forPath(path);
+    stylesheet = Stylesheet.parse(contents, syntax, url: path);
+    visitStylesheet(stylesheet);
+    var results = getMigratedContents();
+    if (results != null) {
+      migrator.migrated[path] = results;
```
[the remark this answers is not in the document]
jathak: Changed this to have `migrated` as a field of `MigrationVisitor`, with any dependency visitors created with `newInstance` getting a reference to the same map so that they can all add to it and the entrypoint's visitor can pass the complete map back to `Migrator` when `run` returns.

Do you think this is fine, or would it be better to have each visitor maintain its own map and then copy the results from each of its dependencies?

lib/src/migrators/module.dart
jathak: I was thinking of the potential case where there are additional use rules, but no patches, but I forgot that if a use rule is added, there would be at least one patch to add a namespace to a reference from that file.

lib/src/migration_visitor.dart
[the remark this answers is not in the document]
nex3: That's a good question. I'd say make the `migrated` field private to `MigrationVisitor` and give it total responsibility for handling the output, but you need to pass it through `newInstance()` somehow...

What if we restructure this so that you don't need to construct multiple visitors at all? Rather than passing a URL to the constructor, pass it to `run()`. That also means that when we get around to de-duplicating repeated imports/uses (which will probably be necessary for large-scale performance), there's a natural place for that as well.

This does mean that per-file state can't just be stored as final fields on the subclass. But you can override `visitStylesheet()` to set it temporarily:

```dart
@override
void visitStylesheet(StylesheetNode node) {
  var oldNamespaces = _namespaces;
  var oldAdditionalUseRules = _additionalUseRules;
  var oldConfigurableVariables = _configurableVariables;
  _namespaces = {};
  _additionalUseRules = Set();
  _configurableVariables = normalizedSet();
  super.visitStylesheet(node);
  _namespaces = oldNamespaces;
  _additionalUseRules = oldAdditionalUseRules;
  _configurableVariables = oldConfigurableVariables;
}
```

This is basically how Dart Sass's `EvaluateVisitor` works. WDYT?

lib/src/migration_visitor.dart
```diff
@@ -28,74 +26,84 @@ import 'utils.dart';
...
       ? Patch.applyAll(patches.first.selection.file, patches)
       : null;
 
-  /// Returns the canonical path of [url] when resolved relative to the current
-  /// path.
-  String resolveImportUrl(String url) =>
-      canonicalizePath(p.join(p.dirname(path), url));
+  /// Resolves [relativeUrl] relative to the current stylesheet's URL.
+  @protected
+  Uri resolveRelativeUrl(Uri relativeUrl) =>
```
nex3: Is this worth making a utility method? I'd probably just inline it.

lib/src/migration_visitor.dart
```diff
@@ -28,74 +26,84 @@ import 'utils.dart';
...
+      stylesheet.span.file.url.resolveUri(relativeUrl);
```
nex3: I prefer referring to a span's URL using [`SourceSpan.sourceUrl`](https://pub.dartlang.org/documentation/source_span/latest/source_span/SourceSpan/sourceUrl.html), since it works for non-`FileSpan`s as well. I think this applies to a couple locations.

lib/src/utils.dart
```diff
@@ -4,21 +4,32 @@
...
 export 'package:sass/src/utils.dart' show normalizedMap, normalizedSet;
 
 import 'patch.dart';
 
-/// Returns the canonical version of [path].
-String canonicalizePath(String path) {
-  return p.canonicalize(resolveImportPath(path));
+/// Returns the canonical version of [url].
+Uri canonicalize(Uri url) {
+  var importer = FilesystemImporter(Directory.current.path);
```
nex3: I'd make this a top-level final variable so you don't have to reinstantiate it over and over—it's stateless so you won't run into problems there. You can also just use `'.'` rather than `Directory.current.path`.

test/migrator_utils.dart
```diff
@@ -44,8 +49,9 @@ _testHrx(File hrxFile, String migrator) async {
             migrated = await MigratorRunner().run(arguments);
           }, getCurrentDirectory: () => Directory(d.sandbox)),
       prints(files.expectedLog?.replaceAll("\$TEST_DIR", d.sandbox) ?? ""));
+  var canonicalize = FilesystemImporter(d.sandbox).canonicalize;
```
nex3: This tear-off usage is kind of non-idiomatic; I'd write

```dart
var importer = FilesystemImporter(d.sandbox);
```

lib/src/utils.dart
jathak: Doing that breaks the tests because we override `Directory.current` to be the sandbox for each one. I'll fix this when I fix #29.

jathak:
Changed `MigrationVisitor` to reuse a single instance for dependencies.

Previously, the module migrator needed to access the `_configurableVariables` of its dependencies' visitors. In the process of changing this, I found #28 and (mostly) fixed this in my changes (though it doesn't yet remove an unnecessary use).

lib/src/migration_visitor.dart
```diff
@@ -28,44 +28,48 @@ import 'utils.dart';
...
   /// The patches to be applied to the stylesheet being migrated.
   UnmodifiableListView<Patch> get patches => UnmodifiableListView(_patches);
-  final List<Patch> _patches = [];
-
-  /// Constructs a new migration visitor, parsing the stylesheet at [url].
-  MigrationVisitor(Uri url, this.migrateDependencies,
-      {Map<Uri, String> migrated})
-      : stylesheet = parseStylesheet(url),
-        this.migrated = migrated ?? {};
+  List<Patch> _patches = [];
```
nex3:
```suggestion
  List<Patch> _patches;
```

The initial value isn't used anyway, since it's always reset in `visitStylesheet()`.

lib/src/migration_visitor.dart
```diff
@@ -78,10 +82,11 @@ abstract class MigrationVisitor extends RecursiveAstVisitor {
       ? Patch.applyAll(patches.first.selection.file, patches)
       : null;
 
-  /// Resolves [relativeUrl] relative to the current stylesheet's URL.
+  /// Adds a new patch that should be applied as part of this migration.
```
nex3: Maybe mention that the patch should be applied to the current stylesheet?

jathak:
The old behavior would be hard to implement with a single MigrationVisitor, so I removed support for configurable variables entirely and will add them back with the new implementation in a separate PR.

lib/src/migrators/module.dart
```diff
@@ -0,0 +1,306 @@
...
+
+  /// Returns the migrated contents of this stylesheet, based on [patches] and
+  /// [_additionalUseRules], or null if the stylesheet does not change.
+  @override
+  String getMigratedContents() {
+    var results = super.getMigratedContents();
+    if (results == null) return null;
+    var semicolon = _currentUrl.path.endsWith('.sass') ? "" : ";";
+    var uses = _additionalUseRules.map((use) => '@use "$use"$semicolon\n');
+    return uses.join("") + results;
```
```suggestion
    return uses.join() + results;
```

lib/src/migrators/module.dart
```diff
@@ -0,0 +1,306 @@
...
+    if (node is Stylesheet) {
+      super.visitChildren(node);
+      return;
+    }
+    _localScope = LocalScope(_localScope);
+    super.visitChildren(node);
+    _localScope = _localScope.parent;
+  }
+
+  /// Adds a namespace to any function call that require it.
```
```suggestion
  /// Adds a namespace to any function call that requires it.
```

lib/src/utils.dart
nex3: Hmm, that's arguably a bug in Dart Sass's API... we shouldn't have an instance of `FilesystemImporter` vary its behavior based on the PWD. This code is fine, though.

PR #17, opened 2019-04-01
Initial implementation of a division migrator

### LINKED ISSUE #20: Add migrator for `/` -> division function
Existing uses of `/` as a division operator should be migrated to the use the new `divide` function (or `math.div` in the new module system)

jathak:
Resolves #20.

This adds a new migrator that migrates the `/` operator to the division function.

By default, this migrator migrates only cases where the `/` is known to be division. In cases the `/` could be division, but whether it is division is not statically known, a warning is emitted instead.

The `--aggressive` flag can be used to migrate cases like `3 / $x + 1`, `3 / $x - 1`, and `fn() / 3` by assuming that `+` and `-` always operate on numbers and that function calls in an expression with `/` always return numbers.

Note: I'm setting the base for this to the `generalization` branch for better diffs, but this should be merged after #16.


lib/src/migrators/division.dart
```diff
@@ -0,0 +1,208 @@
...
+  /// Adds a patch replacing the operator of [node] with ", ".
+  void patchSlashToComma(BinaryOperationExpression node) {
+    var start = node.left.span.end;
+    var end = node.right.span.start;
+    addPatch(Patch(start.file.span(start.offset, end.offset), ", "));
+  }
+
+  /// Runs [operation] with the given context.
+  void withContext(
+      bool isDivisionAllowed, bool expectsNumericResult, void operation()) {
```
nex3:
```suggestion
      void operation(), {bool isDivisionAllowed, bool expectsNumericResult}) {
```

Right now, if I search for `expectsNumericResult` in this file, the calls to `withContext()` that come up are exactly the ones that *don't* modify `_expectsNumericResult`, since they're the ones where it's being passed as an argument. If you make the parameters named and then do e.g.

```dart
  if (isDivisionAllowed != null) _isDivisionAllowed = isDivisionAllowed;
```

you don't have to explicitly pass the old value, *and* the value that's being modified is named at the call site.

lib/src/migrators/division.dart
```diff
@@ -0,0 +1,208 @@
...
+  /// operands.
+  @override
+  void visitBinaryOperationExpression(BinaryOperationExpression node) {
+    if (node.operator == BinaryOperator.dividedBy) {
+      if (shouldMigrate(node)) {
+        addPatch(patchBefore(node, "divide("));
+        patchSlashToComma(node);
+        addPatch(patchAfter(node, ")"));
+      }
+      super.visitBinaryOperationExpression(node);
```
nex3: Shouldn't `_expectsNumericResult` be set here if `shouldMigrate()` returned true?

lib/src/migrators/division.dart
```diff
@@ -0,0 +1,208 @@
...
+          _expectsNumericResult || operatesOnNumbers(node.operator),
+          () => super.visitBinaryOperationExpression(node));
+    }
+  }
+
+  /// Disallows division within this list.
+  @override
+  void visitListExpression(ListExpression node) {
+    withContext(
+        false, _expectsNumericResult, () => super.visitListExpression(node));
```
nex3: `_expectsNumericResult` should also be false here; no matter how the list is used, it won't affect `/` expressions inside it.

lib/src/migrators/division.dart
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  /// If this parenthesized expression contains a division operation, migrates
+  /// it using the parentheses that already exist.
+  @override
+  void visitParenthesizedExpression(ParenthesizedExpression node) {
+    var expression = node.expression;
+    if (expression is BinaryOperationExpression &&
+        expression.operator == BinaryOperator.dividedBy) {
+      withContext(true, _expectsNumericResult, () {
```
nex3: You can avoid some repetition and extra nesting by bringing `withContext()` to the root of the method:

```dart
void visitParenthesizedExpression(ParenthesizedExpression node) {
  withContext(true, _expectsNumericResult, () {
    var expression = node.expression;
    if (expression is BinaryOperationExpression &&
        expression.operator == BinaryOperator.dividedBy &&
        shouldMigrate(expression)) {
      addPatch(patchBefore(node, "divide"));
      patchSlashToComma(expression);
      super.visitBinaryOperationExpression(expression);
    } else {
      super.visitParenthesizedExpression(node);
    }
  });
}
```

lib/src/migrators/division.dart
```diff
@@ -0,0 +1,208 @@
...
+  /// Allows division within this variable declaration.
+  @override
+  void visitVariableDeclaration(VariableDeclaration node) {
+    withContext(true, _expectsNumericResult,
+        () => super.visitVariableDeclaration(node));
+  }
+
+  /// Returns true if we assume that [operator] always returns a number.
+  ///
+  /// This is always true for * and %, and it's true for + and - as long as the
```
nex3:
```suggestion
  /// This is always true for `*` and `%`, and it's true for `+` and `-` as long as the
```

Also below.

lib/src/migrators/division.dart
```diff
@@ -0,0 +1,208 @@
...
+          (operator == BinaryOperator.equals ||
+              operator == BinaryOperator.notEquals);
+
+  /// Returns true if [node] should be treated as division and migrated.
+  ///
+  /// Warns if division is allowed but it's unclear whether or not all types
+  /// are numeric.
+  bool shouldMigrate(BinaryOperationExpression node) {
+    if (!_isDivisionAllowed && onlySlash(node)) {
+      return false;
```
nex3: Nit: I think you can fit this on the previous line.

lib/src/migrators/division.dart
```diff
@@ -0,0 +1,208 @@
...
+    if (node is BinaryOperationExpression) {
+      return node.operator == BinaryOperator.dividedBy &&
+          onlySlash(node.left) &&
+          onlySlash(node.right);
+    }
+    return false;
+  }
+
+  /// Returns true if [node] is believed to always evaluate to a number.
+  bool _allNumeric(Expression node) {
```
nex3: The naming difference between this and `hasNonNumber()` is confusing to me, since they seem like opposites. WDYT about calling them `_isDefinitelyNumber()` and `_isDefinitelyNotNumber()`?

lib/src/migrators/division.dart
```diff
@@ -0,0 +1,208 @@
...
+        addPatch(patchAfter(node, ")"));
+      }
+      super.visitBinaryOperationExpression(node);
+    } else {
+      withContext(
+          true,
+          _expectsNumericResult || operatesOnNumbers(node.operator),
+          () => super.visitBinaryOperationExpression(node));
+    }
+  }
```
nex3: I think `/` operations that aren't migrated to division should be migrated to the `slash-list()` function ([described here](https://github.com/sass/sass/issues/2565#issue-362710498)), at least in aggressive mode. This will ensure that users' runtime behavior won't change when `/` is changed to return a list rather than a string.

test/migrators/division/always_division.hrx
```diff
@@ -0,0 +1,37 @@
...
+  i: 3 / $x < 1;
+}
+
+<==> output/entrypoint.scss
+@function six-divided-by-three() {
+  @return divide(6px, 3px);
+}
+
+a {
+  b: divide((4px + 2px), 3px);
```
nex3: It would be nice, although not necessary, if we could get rid of parentheses here.

lib/src/migrators/division.dart
```diff
@@ -0,0 +1,208 @@
...
+    }
+    return false;
+  }
+
+  /// Returns true if [node] is believed to always evaluate to a number.
+  bool _allNumeric(Expression node) {
+    if (node is NumberExpression) return true;
+    if (node is ParenthesizedExpression) return _allNumeric(node.expression);
+    if (node is UnaryOperationExpression) return _allNumeric(node.operand);
+    if (node is FunctionExpression) return isAggressive;
```
nex3: I think `VariableExpression` should have the same logic as `FunctionExpression` here.

test/migrators/division/always_division.hrx
```diff
@@ -0,0 +1,37 @@
...
+a {
+  b: (4px + 2px) / 3px;
+  c: 6px/3px + 1;
+  d: (6px / 3px);
+  e: 6px / (2px + 1px) / 2;
+  $f: 6px / 3px;
+
+  // * and % always return numbers, so we can assume division here
+  g: (3 * $x) / 4;
+  h: (3 / $x) % 4;
```
nex3:
```suggestion
  h: 3 / $x % 4;
```

Right now, this `/` will be transformed into `divide()` regardless of the operator after it, because of the parentheses.

test/migrators/division/always_division.hrx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+a {
+  b: (4px + 2px) / 3px;
+  c: 6px/3px + 1;
+  d: (6px / 3px);
+  e: 6px / (2px + 1px) / 2;
+  $f: 6px / 3px;
+
+  // * and % always return numbers, so we can assume division here
```
nex3: I think it's worth testing all the operators that force numeric mode in all the positions they can appear.

test/migrators/division/never_division.hrx
```diff
@@ -0,0 +1,9 @@
+<==> input/entrypoint.scss
+a {
+  b: 6px / 3px;
+  c: 6px / 3px / 2;
+  d: #{4px + 2px} / 3px;
+  e: bold 6px/3px sans-serif;
+  f: (bold 6px/3px sans-serif);
+  g: (six / three);
```
nex3: Also test a number and a string (e.g. `6 / foo`) and in an interpolation context (e.g. `:foo(#{1/2}) { ... }`).

lib/src/migrators/division.dart
jathak: Should we do this for the ambiguous slashes that we output warnings for or just the slashes that are unambiguously not division?

lib/src/migrators/division.dart
jathak: I moved `_withContext` to the root of the method, but left the `_shouldMigrate` check as its own if statement, since combining it in this way will cause it to be called a second time on the same expression in `visitBinaryOperationExpression`, which would output duplicate warnings.

lib/src/migrators/division.dart
```diff
@@ -56,128 +56,141 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
...
   }
 
   /// Returns true if [node] is believed to always evaluate to a number.
-  bool _allNumeric(Expression node) {
+  bool _isDefinitelyNumber(Expression node) {
     if (node is NumberExpression) return true;
-    if (node is ParenthesizedExpression) return _allNumeric(node.expression);
-    if (node is UnaryOperationExpression) return _allNumeric(node.operand);
-    if (node is FunctionExpression) return isAggressive;
+    if (node is ParenthesizedExpression) {
```
nex3: Style nit: since these are getting expanded out into multiline `if`s now, I might use `else if` even though they automatically short-circuit just to reduce the number of extra lines.

jathak:
Refactored the patching logic from `visitBinaryOperationExpression` and `visitParenthesizedExpression` into `_visitSlashOperation` to get rid of the non-pure `_shouldMigrate` function and added `slash-list()` support

lib/src/migrators/division.dart
```diff
@@ -90,13 +80,7 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
...
       if (expression is BinaryOperationExpression &&
           expression.operator == BinaryOperator.dividedBy) {
-        if (_shouldMigrate(expression)) {
-          addPatch(patchBefore(node, "divide"));
-          _patchParensIfAny(expression.left);
-          _patchSlashToComma(expression);
-          _patchParensIfAny(expression.right);
-        }
-        super.visitBinaryOperationExpression(expression);
+        _visitSlashOperation(expression, surroundingParens: node);
```
nex3: Rather than passing in `surroundingParens`, which requires `_visitSlashOperation()` to understand a lot about what context it might be called in, consider having this code delete the parentheses itself and just call `_visitSlashOperation(expression)`.

lib/src/migrators/division.dart
```diff
@@ -116,6 +100,65 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
...
+        _isDefinitelyNotNumber(node)) {
+      // Definitely not division
+      if (surroundingParens != null) {
+        addPatch(patchBefore(surroundingParens, "slash-list"));
+      } else {
+        addPatch(patchBefore(node, "slash-list("));
+        addPatch(patchAfter(node, ")"));
+      }
+      _visitSlashListArguments(node);
+    } else if (_expectsNumericResult || _isDefinitelyNumber(node)) {
```
nex3: Consider making `_isDefinitelyNumber()` return `false` for ambiguous binary operations and adding `|| !isPessimistic` to this condition. That way `_isDefinitelyNumber()` returns a result that matches its name regardless of context, and it's more immediately clear when reading this method that a warning will never be emitted in optimistic mode.

lib/src/migrators/division.dart
```diff
@@ -116,6 +100,65 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
...
+      _visitSlashListArguments(node.left);
+      _patchSlashToComma(node);
+      _visitSlashListArguments(node.right);
+    } else if (node is StringExpression &&
+        node.text.contents.length == 1 &&
+        node.text.contents.first is Expression) {
+      var start = node.text.span.start.offset;
+      var end = node.text.span.end.offset;
+      // Remove `#{` and `}`
+      addPatch(Patch(node.span.file.span(start, start + 2), ""));
```
nex3: Consider a `patchDelete()` utility function.

lib/src/migrators/division.dart
```diff
@@ -0,0 +1,266 @@
...
+}
+
+class _DivisionMigrationVisitor extends MigrationVisitor {
+  final bool isPessimistic;
+
+  _DivisionMigrationVisitor(this.isPessimistic, bool migrateDependencies)
+      : super(migrateDependencies: migrateDependencies);
+
+  /// True when division is allowed by the context the current node is in.
+  var _isDivisionAllowed = false;
```
nex3: I just realized, this should also be set to true in argument lists for functions and mixins, except for the specific case of [new-syntax `rgb()` *et al* functions](https://github.com/sass/sass/blob/master/spec/functions/color.md#rgb-and-rgba). Feel free to fix after landing this PR, though.

test/migrators/division/never_division.hrx
```diff
@@ -7,7 +7,17 @@
...
   g: (six / three);
   h: 6 / three;
+  i: #{$x} / #{6px / 3px} / #{2};
 }
 
-<==> log.txt
-Nothing to migrate!
+<==> output/entrypoint.scss
+:foo(#{slash-list(6, 3)}) {
+  b: slash-list(6px, 3px);
```
nex3: Actually, looking at these examples, they're really hard on the eyes. After this lands, can we make the migrator smart enough to only migrate to `slash-list()` in a non-plain-CSS context? Basically, we should generate `slash-list()` only if either `_isDivisionAllowed` is `true` *or* one of the arguments (possibly deeply-nested) is an interpolation expression.

PR #30, opened 2019-04-18
Run tests by invoking the migrator as a subprocess

nex3:
This will make it easier to test a Node.js executable, and avoids
needing a bunch of dependency injection.

### DIFF
--- a/test/migrator_utils.dart
+++ b/test/migrator_utils.dart
@@ -6,23 +6,15 @@
 
 import 'dart:io';
 
-// The sass package's API is not necessarily stable. It is being imported with
-// the Sass team's explicit knowledge and approval. See
-// https://github.com/sass/dart-sass/issues/236.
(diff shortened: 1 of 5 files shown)


test/migrator_utils.dart
```diff
@@ -37,24 +29,33 @@ void testMigrator(String migrator) {
...
+
+  if (files.expectedLog != null) {
+    expect(process.stdout,
+        emitsInOrder(files.expectedLog.trimRight().split("\n")));
   }
+  expect(process.stdout, emitsDone);
+  await process.shouldExit(0);
+
+  await Future.wait(files.output.keys
+      .map((path) => d.file(path, files.output[path]).validate()));
```
jathak: We should also test any input files that aren't in the output to confirm that they are unchanged.

nex3: Good call. Done.

PR #34, opened 2019-04-23
Properly support configurable variables

### LINKED ISSUE #9: Remove variable declarations that are used to configure an import
At some point, it would be good to remove variable declarations that are used to configure an import. Maybe file an issue for that?

_Originally posted by @nex3 in https://github.com/sass/module_migrator/pull/4_

### LINKED ISSUE #28: Handle case where a stylesheet is configuring an indirect dependency
Something like this

```scss
// entrypoint.scss
$config: red;
@import "direct";

// _direct.scss
@import "indirect";

// _indirect.scss
$config: green !default;
```

is currently migrated to

```scss
// entrypoint.scss
$config: red;
@use "direct";

// _direct.scss
@use "entrypoint";
@use  "indirect";

// _indirect.scss
$config: green !default;
```

which is definitely wrong, but I'm also not sure what the best way to actually do it would be.

jathak:
Resolves #9 and resolves #28.

This re-adds support for migrating configurable variables with a new implementation that stores all global declarations of a variable in the order they're and then uses that information to determine the configuration necessary for each `@use` rule.

I'm not super happy with the way this code [inserts fake declarations](https://github.com/sass/migrator/pull/34/files#diff-c89554bdef9ec7c390f3bdc5ecbf4188R244) into the global variables to keep track of forwarded config variables, but I'm not sure what a cleaner alternative would be.


test/migrators/module/indirect_configuration.hrx
```diff
@@ -0,0 +1,22 @@
...
+@import "indirect";
+
+<==> input/_indirect.scss
+$config: green !default;
+
+<==> output/entrypoint.scss
+$config: red;
+@use "direct" with (
+  $config: red
+);
```
nex3: It would be nice to generate this as `@use "direct" with ($config: red);` for the case where only a single variable is being configured.

test/migrators/module/indirect_configuration.hrx
```diff
@@ -0,0 +1,22 @@
...
+@import "direct";
+
+<==> input/_direct.scss
+@import "indirect";
+
+<==> input/_indirect.scss
+$config: green !default;
+
+<==> output/entrypoint.scss
+$config: red;
```
nex3: We should get rid of the original variable definitions if they're used for configuration, and replace later uses with `$direct.config`.

test/migrators/module/indirect_configuration_with_reference.hrx
```diff
@@ -0,0 +1,30 @@
...
+
+<==> input/entrypoint.scss
+$config: red;
+@import "direct";
+
+<==> input/_direct.scss
+@import "indirect";
+
+a {
+  b: $config;
```
nex3: Also test a reference that appears in the same file as the override.

lib/src/migrators/module.dart
```diff
@@ -43,14 +43,17 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+  final _globalVariables = normalizedMap<VariableDeclaration>();
 
   /// Global mixins defined at any time during the migrator run.
   final _globalMixins = normalizedMap<MixinRule>();
 
   /// Global functions defined at any time during the migrator run.
   final _globalFunctions = normalizedMap<FunctionRule>();
 
+  /// Stores whether a given VariableDeclaration has been referenced.
+  final _variableReferenced = <VariableDeclaration, bool>{};
```
```suggestion
  final _referencedVariables = Set<VariableDeclaration>();
```

Maps with boolean keys are just Sets with weirder APIs, imo.

lib/src/migrators/module.dart
```diff
@@ -43,14 +43,17 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
-  final _globalVariables = normalizedMap<List<VariableDeclaration>>();
+  final _globalVariables = normalizedMap<VariableDeclaration>();
 
   /// Global mixins defined at any time during the migrator run.
   final _globalMixins = normalizedMap<MixinRule>();
 
   /// Global functions defined at any time during the migrator run.
   final _globalFunctions = normalizedMap<FunctionRule>();
 
+  /// Stores whether a given VariableDeclaration has been referenced.
```
nex3: Define exactly what "referenced" means. Is it referring to it in an expression? Reassigning it? Both?

lib/src/migrators/module.dart
```diff
@@ -79,14 +86,17 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
   /// the module migrator will filter out the dependencies' migration results.
   _ModuleMigrationVisitor() : super(migrateDependencies: true);
 
+  String get semicolonIfNotIndented =>
```
nex3: Nit: document this and make it private.

PR #41, opened 2019-05-14
Improvements to the division migrator

### LINKED ISSUE #32: Treat slash as division in argument lists
I just realized, this should also be set to true in argument lists for functions and mixins, except for the specific case of [new-syntax `rgb()` *et al* functions](https://github.com/sass/sass/blob/master/spec/functions/color.md#rgb-and-rgba). Feel free to fix after landing this PR, though.

_Originally posted by @nex3 in https://github.com/sass/migrator/pull/17_

### LINKED ISSUE #33: Only migrate to slash-list in a non-plain-CSS context
Actually, looking at these examples, they're really hard on the eyes. After this lands, can we make the migrator smart enough to only migrate to `slash-list()` in a non-plain-CSS context? Basically, we should generate `slash-list()` only if either `_isDivisionAllowed` is `true` *or* one of the arguments (possibly deeply-nested) is an interpolation expression.

_Originally posted by @nex3 in https://github.com/sass/migrator/pull/17_

jathak:
Resolves #32 and resolves #33.


lib/src/migrators/division.dart
```diff
@@ -65,6 +73,49 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
...
 
+  /// Allows division within a function call's arguments, with special handling
+  /// for new-syntax color functions.
+  @override
+  void visitFunctionExpression(FunctionExpression node) {
+    visitInterpolation(node.name);
+    if (node.name.asPlain == "rgb" ||
+        node.name.asPlain == "rgba" ||
+        node.name.asPlain == "hsl" ||
+        node.name.asPlain == "hsla") {
```
nex3: It would be great to remove some of this nesting... maybe factor this out into `bool _tryColorFunction(FunctionExpression node)` that can just `return false` every time it determines that the function doesn't need special handling?

lib/src/migrators/division.dart
```diff
@@ -65,6 +73,49 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
...
+          node.arguments.named.length == 1) {
+        channels = node.arguments.named[r'$channels'];
+      }
+      if (channels != null &&
+          channels is ListExpression &&
+          !channels.hasBrackets &&
+          channels.separator == ListSeparator.space &&
+          channels.contents.length == 3 &&
+          channels.contents[2] is BinaryOperationExpression) {
+        var red = channels.contents[0];
```
nex3: Using `red` *et al* is a little confusing, because this same path is used for `hsl()` colors. Maybe `channel1` *et al* instead?

lib/src/migrators/division.dart
```diff
@@ -115,9 +166,11 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
     if ((!_isDivisionAllowed && _onlySlash(node)) ||
         _isDefinitelyNotNumber(node)) {
       // Definitely not division
-      addPatch(patchBefore(node, "slash-list("));
-      addPatch(patchAfter(node, ")"));
-      _visitSlashListArguments(node);
+      if (_isDivisionAllowed || _containsInterpolation(node)) {
```
nex3: Maybe add a comment here about why we don't always want to use `slash-list()`.

test/migrators/division/always_division.hrx
```diff
@@ -49,4 +64,9 @@ a {
   l: divide(3, $x) > 1;
   m: divide(3, $x) <= 1;
   n: divide(3, $x) >= 1;
+
+  // Function calls
+  o: identity(divide(6px, 3px));
+  p: rgba(10, 20, divide(30, 2), 0.5);
+  q: rgb(10 20 divide(30, 2) / 0.5);
```
nex3: This actually isn't a valid translation; because `divide()` isn't a literal number, this `/` is *always* interpreted as division, which means the metadata about what the right-hand side used to be gets lost before it reaches `rgb()`. (This is the kind of nasty confusing edge case that makes the current `/` semantics so awful.)

I think the ideal translation here is probably `rgb(10, 20, divide(30, 2), 0.5)`, but it's also reasonable to do `rgb(10 20 slash-list(divide(30, 2), 0.5))` if that's easier.

test/migrators/division/never_division.hrx
```diff
@@ -8,16 +8,18 @@
...
   i: #{$x} / #{6px / 3px} / #{2};
+  j: rgb(10 20 30 / 0.5);
 }
 
 <==> output/entrypoint.scss
-:foo(#{slash-list(6, 3)}) {
-  b: slash-list(6px, 3px);
-  c: slash-list(6px, 3px, 2);
+:foo(#{6/3}) {
+  b: 6px / 3px;
```
nex3: Now that these are being preserved as `/`, it's probably a good idea to add additional test cases where they're assigned to variables so we can verify that they're converted to `slash-list()` when necessary.

lib/src/migrators/division.dart
```diff
@@ -156,6 +122,44 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
         isDivisionAllowed: true);
   }
 
+  /// Migrates [node] and returns true if it is a new-syntax color function or
+  /// returns false if it is any other function.
+  bool _tryColorFunction(FunctionExpression node) {
+    if (!["rgb", "rgba", "hsl", "hsla"].contains(node.name.asPlain)) {
+      return false;
+    }
```
nex3: Consider adding empty lines after each `return false` to visually emphasize each distinct check.

lib/src/migrators/division.dart
```diff
@@ -156,6 +122,44 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
...
+    if (channels == null ||
+        channels.hasBrackets ||
+        channels.separator != ListSeparator.space ||
+        channels.contents.length != 3 ||
+        channels.contents.last is! BinaryOperationExpression) {
+      return false;
+    }
+    var last = channels.contents.last as BinaryOperationExpression;
+    if (last.left is! NumberExpression || last.right is! NumberExpression) {
+      _patchSpacesToCommas(channels);
```
nex3: It's probably a good idea to provide an example of why this is necessary here. A really dedicated reader who's wondering what this is for could remove this code and see what tests fail, but it's better to save them the trouble.

nex3:
Great work!

nex3:
Oh, one other thing I forgot to mention: now that we're doing alpha releases, anything that changes the user-visible API surface of the package should get at least a mention in the CHANGELOG.

Issue #21, opened 2019-04-03
Option to unprefix members when migrating a library to the module system


jathak:
(Part of #15)

The user should be able to specify a prefix to remove from all members when migrating a library to the new module system. The migrator should also generate a `<library>.import.scss` file that forwards all members with the prefix to allow for migrating a library without breaking downstream dependencies.

PR #52, opened 2019-05-31
Fixes for shadowing/reassignment of variables in other modules

jathak:
Found a bug in local variable shadowing while working on #21.


jathak:
Travis is failing due to #53.

test/migrators/module/shadowed_variables.hrx
```diff
@@ -0,0 +1,42 @@
+<==> arguments
+--migrate-deps
+
+<==> input/entrypoint.scss
+@import "library";
+$a : red;
```
nex3:
```suggestion
$a: red;
```

test/migrators/module/shadowed_variables.hrx
```diff
@@ -0,0 +1,42 @@
...
+
+<==> input/_library.scss
+$a: blue;
+$b: blue;
+$c: blue;
+$d: blue;
+
+<==> output/entrypoint.scss
+@use "library";
+$a : red;
```
nex3: Why does this shadow the library's definition rather than assigning to it?

jathak: Is `$library.a: red;` valid?

jathak: Ping

nex3: Oh, sorry! Yes, it is.

test/migrators/module/shadowed_variables.hrx
```diff
@@ -0,0 +1,42 @@
...
+
+<==> input/_library.scss
+$a: blue;
+$b: blue;
+$c: blue;
+$d: blue;
+
+<==> output/entrypoint.scss
+@use "library";
+$a: red;
```
nex3: I'm confused... didn't we decide this should be `$library.a: red;`?

jathak: Oh, sorry! Misread that. I'll change this then.

jathak:
Okay, now reassignment of global variables from another file should reassign to that module's variable instead of declaring a new one in the current module, as long as the original declaration did not come from an upstream file (to prevent cases like the one in [`configuration.hrx`](https://github.com/sass/migrator/blob/master/test/migrators/module/configuration.hrx) trying to namespace `$no-config` in `_library.scss` as `$entrypoint.no-config`).

lib/src/migrators/module.dart
```diff
@@ -59,6 +59,13 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
   /// expression after being declared.
   final _referencedVariables = <VariableDeclaration>{};
 
+  /// List of stylesheets currently being migrated.
+  ///
+  /// Used to ensure that a dependency declaring a variable that an upstream
+  /// stylesheet already declared is not treated as reassignment (since that
+  /// would cause a circular dependency).
+  final _upstreamStylesheets = <Uri>[];
```
nex3: Since you're calling `.contains()` on this a lot, it should probably be a `Set`.

lib/src/migrators/module.dart
```diff
@@ -389,10 +398,23 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
     if (_localScope == null || node.isGlobal) {
-      if (node.isGuarded &&
-          _globalVariables.containsKey(node.name) &&
+      if (_globalVariables.containsKey(node.name) &&
           _globalVariables[node.name].span.sourceUrl != _currentUrl) {
-        _configuredVariables.add(_globalVariables[node.name]);
+        if (node.isGuarded) {
+          _configuredVariables.add(_globalVariables[node.name]);
+        } else if (!_upstreamStylesheets
+            .contains(_globalVariables[node.name].span.sourceUrl)) {
```
nex3: Consider putting `_globalVariables[node.name].span.sourceUrl` (and maybe also `_globalVariables[node.name]`) in a variable since it's long and referenced multiple times.

lib/src/migrators/module.dart
```diff
@@ -389,10 +398,23 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+          // This declaration reassigns a variable in another module. Since we
+          // don't care about the actual value of the variable while migrating,
+          // we leave the node in _globalVariables as-is, so that future
+          // references namespace based on the original declaration, not this
+          // reassignment.
```
nex3: Nice comment that clearly explains why this code isn't doing something you might expect it to do :+1:.

PR #61, opened 2019-06-10
Emit warnings to stderr

### LINKED ISSUE #14: Warnings should be logged to stderr
Re: https://github.com/sass/module_migrator/pull/13#discussion_r264477889

This will require changing the tests to check stderr instead of printed output, which will probably require using test_process.

jathak:
Resolves #14.

This also renames the `warn` function in `lib/src/utils.dart` to `emitWarning` to avoid conflicting with the `warn` function now exposed by Sass.

### DIFF
--- a/lib/src/io/node.dart
+++ b/lib/src/io/node.dart
@@ -1,0 +1,20 @@
+// Copyright 2019 Google LLC
+//
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import 'package:js/js.dart';
+
+export 'package:node_io/node_io.dart';
+
+// ignore: missing_js_lib_annotation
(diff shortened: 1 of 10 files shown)


lib/src/io/node.dart
```diff
@@ -0,0 +1,20 @@
...
+
+import 'package:js/js.dart';
+
+export 'package:node_io/node_io.dart';
+
+// ignore: missing_js_lib_annotation
+@JS("process.stderr.write")
+external _writeToStderr(String text);
+
+class Stderr {
```
nex3: This seems like it should be in `node_io`. If it's not there yet, can you send them a PR?

jathak: I took a look at this, but implementing stderr properly there (rather than just what's needed here) is more of a yak shave than I'm up for at the moment, so I'll close this for now in favor of #62 to fix the compile error.

PR #66, opened 2019-07-10
Add tests for indented syntax

### LINKED ISSUE #5: Support indented syntax
```suggestion
    print('Migrates one or more Sass files to the new module system.\n\n'
```

This should ultimately work with the indented syntax as well as SCSS.

_Originally posted by @nex3 in https://github.com/sass/module_migrator/pull/4/files_

jathak:
Resolves #5.

This adds copies of the unprefix and namespace_mixin tests that use the indented syntax and the shorthand `+`/`=` syntax for mixins.

I initially thought about making indented syntax copies of all tests, but that felt redundant, as these two tests should have enough to test the basics of the indented syntax and the one special case I could find (the mixin shorthand syntax).

Are there any other things particular to the indented syntax that would be good to add tests for or do these seem fine?

### DIFF
--- a/test/migrators/module/namespace_mixin.hrx
+++ b/test/migrators/module/namespace_mixin.hrx
@@ -1,6 +1,9 @@
 <==> arguments
 --migrate-deps
 
+<==> README.md
+This test should be kept in sync with namespace_mixin_indented.hrx.
+
 <==> input/entrypoint.scss
 @import "library";
 a {
--- a/test/migrators/module/namespace_mixin_indented.hrx
+++ b/test/migrators/module/namespace_mixin_indented.hrx
@@ -4,6 +4,9 @@
 <==> README.md
 This test should be kept in sync with namespace_mixin.hrx.
 
+This test is duplicated for the indented syntax since it includes the shorthand
+mixin syntax that's not present in the SCSS syntax.
+
 <==> input/entrypoint.sass
 @import "library"
 a
--- a/test/migrators/module/namespace_mixin_indented.hrx
+++ b/test/migrators/module/namespace_mixin_indented.hrx
@@ -1,0 +1,21 @@
+<==> arguments
+--migrate-deps
+
+<==> README.md
+This test should be kept in sync with namespace_mixin.hrx.
+
+<==> input/entrypoint.sass
+@import "library"
+a
+  +helper
+  @include helper
+
+<==> input/_library.sass
+=helper
+  display: block
+
+<==> output/entrypoint.sass
+@use "library"
+a
+  +library.helper
+  @include library.helper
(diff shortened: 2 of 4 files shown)


jathak:
Hmm. Not sure why the Node tests are failing on Travis. They pass on my machine

nex3:
It's probably a good idea to add comments to these tests indicating why they were chosen in particular.

> Hmm. Not sure why the Node tests are failing on Travis. They pass on my machine

Are you using the same Node and Dart versions? Usually that sort of thing is caused by version skew.

jathak:
Looks like running `pub upgrade` caused the Node tests to start failing on my machine, so my guess is that something between sass 1.21.0 and sass 1.22.3 broke them.

From the stack trace it looks like it's some sort of type error, but I'm not sure where exactly it's coming from. Is there an easy way to convert the JS stack trace to a Dart one with the dart2js source map?

nex3:
You can try to do something manually with https://pub.dev/packages/source_map_stack_trace, but I tend to find it easier to just debug into the generated JS.

jathak:
Okay, so it seems like the error is occuring when the migrator calls `FilesystemImporter.canonicalize` on the entrypoint path that's passed in. Not sure why yet

jathak:
Okay, so I looks like [this cast](https://github.com/sass/dart-sass/blame/db87ed2f0efc1fb40936fa55ac310e5728ae9510/lib/src/io/node.dart#L164) is what's causing the exception, with the error message being:
```
CastError: Instance of 'NullError': type 'NullError' is not a subtype of type '_SystemError'
```

`canonicalize` is checking whether `entrypoint.import.scss` exists, which it doesn't, but the exception that the Node function throws isn't the expected `_SystemError` for some reason.

jathak:
Relevant compiled JS:
```js
    interceptedTypeCast: function(value, property) {
      var t1;
      if (value != null)
        t1 = (typeof value === "object" || typeof value === "function") && J.getInterceptor$(value)[property];
      else
        t1 = true;
      if (t1)
        return value;
      H.propertyTypeCastError(value, property);
    },
...
    fileExists: function(path) {
      var error, systemError, t1, exception;
      try {
        t1 = J.isFile$0$x(J.statSync$1$x(self.fs, path));
        return t1;
      } catch (exception) {
        error = H.unwrapException(exception);
        systemError = H.interceptedTypeCast(error, "$is_SystemError");
        if (J.$eq$(J.get$code$x(systemError), "ENOENT"))
          return false;
        throw exception;
      }
    },
```

nex3:
I assume the `path` that's getting passed in is `null`, which it shouldn't be. What's the call stack for the call to `fileExists()`?

jathak:
I figured it out. `_fs` is null when we try to call `_fs.statSync(path).isFile()`.

Could this be related to sass/dart-sass#727?

jathak:
#67 should fix this

PR #69, opened 2019-07-17
Migrate nested imports to load-css calls

### LINKED ISSUE #10: Migrate nested imports
Nested `@import`s should be replaced with `@include meta.load-css()`.

jathak:
Resolves #10.

As part of this, I refactored `LocalScope` into `Scope`, including both local and global members.

All `@import` rules that can't be migrated to `@use` will instead be migrated to `@include meta.load-css`. This includes both nested imports and top-level imports after a rule other than `@charset`, `@use`, `@forward`, or `@import`.


lib/src/migrators/module/scope.dart
```diff
@@ -0,0 +1,140 @@
...
+    while (current != null) {
+      flattened.insertFrom(current);
+      current = current.parent;
+    }
+    return flattened;
+  }
+
+  /// Inserts all direct members of [other] into this scope, excluding any members
+  /// that also existing in [excluding].
+  void insertFrom(Scope other, {Scope excluding}) {
```
nex3: This name is a little hard to understand... maybe `addAllMembers`?

lib/src/migrators/module.dart
```diff
@@ -343,6 +345,10 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     _configuredVariables = oldConfiguredVariables;
 
     if (externallyConfiguredVariables.isNotEmpty) {
+      if (migrateToLoadCss) {
+        // TODO(jathak): Make this error message meaningful
```
nex3: Can you do this in this PR?

lib/src/migrators/module.dart
```diff
@@ -361,32 +367,45 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
       } else {
         // TODO(jathak): Handle the case where the expression of this
         // declaration has already been patched.
-        addPatch(patchDelete(variable.span));
+        addPatch(
+            patchDelete(variable.span, start: -variable.span.start.column));
```
nex3: It's unlikely, but this could break if the variable is on a line with other code. Maybe teach `patchDelete()` to delete the entire line if all that's left is whitespace?

lib/src/migrators/module.dart
```diff
@@ -438,21 +459,22 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
   /// Declares a variable within this stylesheet, in the current local scope if
   /// it exists, or as a global variable otherwise.
   void _declareVariable(VariableDeclaration node) {
-    if (_localScope == null || node.isGlobal) {
-      var name = _unprefix(node.name);
+    var name = node.name;
+    if (_scope.parent == null || node.isGlobal) {
```
nex3: Consider making `_scope.parent == null` into a getter, such as `_scope.isGlobal`, to better communicate the semantics of what you're checking.

lib/src/migrators/module.dart
```diff
@@ -527,4 +545,88 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+  @override
+  void visitAtRootRule(AtRootRule node) {
+    _useAllowed = false;
+    super.visitAtRootRule(node);
+  }
+
+  /// Disallows @use after at-rules other than @charset.
+  @override
+  void visitAtRule(AtRule node) {
+    if (node.name.asPlain == 'charset') _useAllowed = false;
```
nex3: `@charset` rules are discarded during parse-time, so you don't need to worry about them here.

lib/src/migrators/module/scope.dart
```diff
@@ -0,0 +1,140 @@
...
+  final functions = normalizedMap<FunctionRule>();
+
+  /// Nested imports from this scope.
+  ///
+  /// Note: For the purposes of migration, any imports that come after a rule
+  /// other than @import, @use, @forward, or @charset are considered nested, since
+  /// they can't be migrated to @use.
+  final nestedImports = <Uri>{};
+
+  Scope(this.parent);
```
nex3:
```suggestion
  Scope([this.parent]);
```

Making `parent` optional helps communicate that constructing a parentless `Scope` is valid and expected, and it makes the callsite where that happens cleaner.

lib/src/migrators/module/scope.dart
```diff
@@ -0,0 +1,140 @@
...
+      parent != null && (mixins.containsKey(name) || parent.isLocalMixin(name));
+
+  /// Returns whether a function [name] exists in a non-global scope.
+  bool isLocalFunction(String name) =>
+      parent != null &&
+      (functions.containsKey(name) || parent.isLocalFunction(name));
+
+  /// Constructs and throws a MigrationException for a reference to a member
+  /// from a nested import.
+  void _errorNestedImport(String type, Uri importUrl, FileSpan reference) {
```
nex3: Explain this in more detail. We can expect that most users of the migrator aren't going to be intimately familiar with the module system, so they may not understand that nested imports get translated to the `load-css()` mixin.

Maybe something like "This $type was loaded from a nested import. The module system only supports loading nested CSS using the load-css() function, which doesn't load $types."

lib/src/migrators/module/scope.dart
```diff
@@ -0,0 +1,140 @@
...
+  /// throwing a MigrationException if it does.
+  void checkNestedImportVariable(VariableExpression node) {
+    if (variables.containsKey(node.name)) {
+      var url = variables[node.name].span.sourceUrl;
+      if (nestedImports.contains(url)) {
+        _errorNestedImport('variable', url, node.span);
+      }
+    } else {
+      parent?.checkNestedImportVariable(node);
+    }
```
nex3: Can you factor out the implementations of these functions, maybe by passing in an `AstNode Function(Scope)` to abstract over `variables[node.name]` *et al*?

lib/src/migrators/module/scope.dart
```diff
@@ -0,0 +1,140 @@
...
+      parent?.checkNestedImportFunction(node);
+    }
+  }
+
+  /// Create a flattened version of this scope, combining all members from all
+  /// ancestors into one level.
+  ///
+  /// This is used for migrating nested imports, allowing the migrator to treat
+  /// local members from the upstream stylesheet as global ones while migrating
+  /// the nested import.
```
nex3: Is this desirable? Won't it produce an incorrect migration if a local member is referred to by the nested import? (That's a case that *can't* be migrated, so I'd expect an error.)

jathak:
Made changes based on your comments.

Flattening the scope for the nested import is mostly there to allow local variables to be used for configuration. I've changed it now to keep track of specific members that are unreferencable (both global members from the nested import in the outer stylesheet and local members from the outer stylesheet in the nested import) so that local variables can still be used for configuration, but they can't otherwise be referenced.

lib/src/migrators/module.dart
```diff
@@ -317,23 +318,28 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
 
     var oldScope = _scope;
     if (migrateToLoadCss) {
-      _scope = _scope.flatten();
+      _scope = oldScope.copyForNestedImport();
+      var current = oldScope.parent;
```
nex3: Maybe add a comment here explaining that variables are special because they may be used to configure the nested import.

lib/src/migrators/module.dart
```diff
@@ -346,8 +352,12 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
 
     if (externallyConfiguredVariables.isNotEmpty) {
       if (migrateToLoadCss) {
-        // TODO(jathak): Make this error message meaningful
-        throw MigrationException("Invalid nested import");
+        var firstConfig = externallyConfiguredVariables.values.first;
+        throw MigrationException(
+            "This declaration attempts to override a default value in an "
+            "indirect, nested import of ${p.prettyUri(_lastUrl)}, which is "
+            "not possible in the module system.",
```
nex3: Very nice :smile:.

lib/src/migrators/module/scope.dart
```diff
@@ -14,6 +14,8 @@ import 'package:sass/src/ast/sass.dart';
 import 'package:path/path.dart' as p;
 import 'package:source_span/source_span.dart';
 
+enum UnreferencableType { localFromImporter, globalFromNestedImport }
```
nex3: It's generally good style to use pseudo-enums (such as Dart Sass's [`Syntax`](https://github.com/sass/dart-sass/blob/master/lib/src/syntax.dart)) rather than Dart enums, because they're easier to give extra behavior to and control the stringification of.

lib/src/migrators/module/scope.dart
```diff
@@ -58,83 +65,90 @@ class Scope {
...
+          throw MigrationException(
+              "This $type was loaded from a nested import of $uri. The module "
+              "system only supports loading nested CSS using the load-css() "
+              "mixin, which doesn't load ${type}s.",
+              span: reference);
+        }
       }
     } else {
-      parent?.checkNestedImportVariable(node);
+      parent?._checkUnreferencable(type, reference, findDeclaration);
```
nex3: Consider putting this clause first and returning early to avoid extra indentation:

```dart
if (declaration == null) {
  parent?._checkUnreferencable(type, reference, findDeclaration);
  return;
}
if (!unreferencableMembers.containsKey(declaration)) return;

// ...
```

lib/src/migrators/module/scope.dart
```diff
@@ -58,83 +65,90 @@ class Scope {
...
+              "load-css() mixin, which doesn't allow access to local ${type}s "
+              "from the outer stylesheet.",
+              span: reference);
+        } else if (unrefType == UnreferencableType.globalFromNestedImport) {
+          throw MigrationException(
+              "This $type was loaded from a nested import of $uri. The module "
+              "system only supports loading nested CSS using the load-css() "
+              "mixin, which doesn't load ${type}s.",
+              span: reference);
+        }
```
nex3: Throw a `StateError` here about an unrecognized type, just for completeness's sake.

PR #70, opened 2019-07-22
Don't migrate CSS-compatiblity overloads

### LINKED ISSUE #54: Don't module-ify CSS-compatibility overloads
A few function overloads exist to provide compatibility with plain CSS function calls, and should not be migrated to the new module system. These are:

* `grayscale()` with a number argument.
* `invert()` with a number argument.
* `opacity()` with a number argument.
* `alpha()` with multiple arguments, or with a single argument that's a single-equals expression.
* `saturate()` with a single argument.

See also https://github.com/sass/sass/issues/2703.

jathak:
Resolves #54.

### DIFF
--- a/lib/src/migrators/module.dart
+++ b/lib/src/migrators/module.dart
@@ -241,6 +241,18 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
         : null;
 
     if (namespace == null && builtInFunctionModules.containsKey(name)) {
+      // Don't migrate CSS-compatibility overloads.
+      var argument = getOnlyArgument(node.arguments);
+      if (({'grayscale', 'invert', 'opacity'}.contains(name) &&
+              argument is NumberExpression) ||
+          (name == 'saturate' && argument != null) ||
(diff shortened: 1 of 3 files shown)


lib/src/migrators/module.dart
```diff
@@ -241,6 +241,18 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+      // Don't migrate CSS-compatibility overloads.
+      var argument = getOnlyArgument(node.arguments);
+      if (({'grayscale', 'invert', 'opacity'}.contains(name) &&
+              argument is NumberExpression) ||
+          (name == 'saturate' && argument != null) ||
+          (name == 'alpha' &&
+              (node.arguments.positional.length + node.arguments.named.length >
+                      1 ||
+                  (argument is BinaryOperationExpression &&
+                      argument.operator == BinaryOperator.singleEquals)))) {
```
nex3: Rather than having this be a big complex boolean expression, maybe factor it out in a function so that you can do short-circuit returns?

PR #72, opened 2019-08-01
Add flag for forwarding members from entrypoint

### LINKED ISSUE #47: Add a flag to forward all members from the entrypoint
When running the migrator on a library, authors will usually want to expose all the members that library defines from the library's entrypoint file (that is, the file passed initially to the migrator). The migrator should support a flag to do this automatically by `@forward`ing all the files that define public names that are transitively imported by the migrator.

jathak:
Resolves #47.

Adds a new flag `--forward` which can be `none` (default), `all`, or `prefixed`.

If `--forward=all`, all global members from dependencies will be forwarded from the entrypoint.

If `--forward=prefixed`, only members that were previously prefixed with the value of `--remove-prefix` will be forwarded.

### DIFF
--- a/lib/src/migrators/module.dart
+++ b/lib/src/migrators/module.dart
@@ -12,15 +12,15 @@ import 'package:args/args.dart';
 import 'package:sass/src/ast/sass.dart';
 
 import 'package:path/path.dart' as p;
-import 'package:sass_migrator/src/migrators/module/forward_type.dart';
 import 'package:source_span/source_span.dart';
 
-import 'package:sass_migrator/src/migration_visitor.dart';
-import 'package:sass_migrator/src/migrator.dart';
-import 'package:sass_migrator/src/patch.dart';
-import 'package:sass_migrator/src/utils.dart';
+import '../migration_visitor.dart';
+import '../migrator.dart';
+import '../patch.dart';
+import '../utils.dart';
 
 import 'module/built_in_functions.dart';
+import 'module/forward_type.dart';
 import 'module/scope.dart';
 import 'module/unreferencable_type.dart';
 
--- a/lib/src/migrators/module.dart
+++ b/lib/src/migrators/module.dart
@@ -12,6 +12,7 @@ import 'package:args/args.dart';
 import 'package:sass/src/ast/sass.dart';
 
 import 'package:path/path.dart' as p;
+import 'package:sass_migrator/src/migrators/module/forward_type.dart';
 import 'package:source_span/source_span.dart';
 
 import 'package:sass_migrator/src/migration_visitor.dart';
@@ -54,15 +55,17 @@ class ModuleMigrator extends Migrator {
   /// If [migrateDependencies] is false, the migrator will still be run on
   /// dependencies, but they will be excluded from the resulting map.
   Map<Uri, String> migrateFile(Uri entrypoint) {
-    if (argResults['forward'] == 'prefixed' &&
+    var forward = ForwardType(argResults['forward']);
+    if (forward == ForwardType.prefixed &&
         argResults['remove-prefix'] == null) {
       throw MigrationException(
-          'Error: Must provide --remove-prefix when --forward=prefixed');
+          'You must provide --remove-prefix with --forward=prefixed so we know '
+          'which prefixed members to forward.');
     }
     var migrated = _ModuleMigrationVisitor(
             prefixToRemove:
                 (argResults['remove-prefix'] as String)?.replaceAll('_', '-'),
-            forwardOption: argResults['forward'] as String)
+            forward: forward)
         .run(entrypoint);
     if (!migrateDependencies) {
       migrated.removeWhere((url, contents) => url != entrypoint);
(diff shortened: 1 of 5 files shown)


lib/src/migrators/module.dart
```diff
@@ -42,9 +54,15 @@ class ModuleMigrator extends Migrator {
   /// If [migrateDependencies] is false, the migrator will still be run on
   /// dependencies, but they will be excluded from the resulting map.
   Map<Uri, String> migrateFile(Uri entrypoint) {
+    if (argResults['forward'] == 'prefixed' &&
+        argResults['remove-prefix'] == null) {
+      throw MigrationException(
+          'Error: Must provide --remove-prefix when --forward=prefixed');
```
nex3: I might make this a little clearer about *why* this is necessary. Maybe "You must provide --remove-prefix with --forward=prefixed so we know which prefixed members to forward" or something like that?

Also, I'd avoid putting "Error: " at the beginning of literal exception messages. If you want a header like that, it should be the responsibility of the error handler that catches and prints the exceptions.

lib/src/migrators/module.dart
```diff
@@ -96,12 +114,15 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
   /// should be removed.
   final String prefixToRemove;
 
+  /// The value of the --forward flag, either 'none', 'all', or 'prefixed'.
+  final String forwardOption;
```
nex3: "Option" here is kind of redundant. I'd probably just call it `forward`.

Also, I'd translate the string value to an enum at some point before it gets here. It's more overhead, but it gives you type-checking for any checks you do in the body of this visitor.

lib/src/migrators/module.dart
```diff
@@ -117,7 +138,86 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     if (results == null) return null;
     var uses = _additionalUseRules
         .map((use) => '@use "$use"$_semicolonIfNotIndented\n');
-    return uses.join() + results;
+    return _getEntrypointForwards() + uses.join() + results;
+  }
+
+  /// Returns whether an member of [name] should be forwarded in the entrypoint.
```
```suggestion
  /// Returns whether the member named [name] should be forwarded in the entrypoint.
```

lib/src/migrators/module.dart
```diff
@@ -117,7 +138,86 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+    // be forwarded.
+    var forwards = <String>[];
+    for (var url in shown.keys) {
+      var shownCount = shown[url].length;
+      var hiddenCount = hidden[url]?.length ?? 0;
+      var forward = '@forward "${_absoluteUriToDependency(url)}"';
+
+      // When not all members from a dependency should be forwarded, determine
+      // whether to use a `show` clause or a `hide` clause based on the number
+      // of members in each set.
```
nex3: Nice! I like how this avoids `show`/`hide` declarations when they aren't necessary.

lib/src/migrators/module.dart
```diff
@@ -117,7 +138,86 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+      if (hiddenCount > 0) {
+        if (shownCount <= hiddenCount) {
```
nex3: This is clever, but I think we probably want to decide which style we want to encourage and use that universally. I lean towards defaulting to `hide`.

lib/src/migrators/module.dart
```diff
@@ -578,17 +678,23 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
-      if (basename.startsWith('_')) basename = basename.substring(1);
-      var simplePath = p.relative(p.join(p.dirname(relativePath), basename));
+      var simplePath = _absoluteUriToDependency(node.span.sourceUrl);
       _additionalUseRules.add(simplePath);
       _namespaces[node.span.sourceUrl] = namespaceForPath(simplePath);
     }
     return _namespaces[node.span.sourceUrl];
   }
 
+  /// Converts an absolute URI for a stylesheet into the simplest string that
```
```suggestion
  /// Converts an absolute URL for a stylesheet into the simplest string that
```

Also in the function name. This is kind of a personal crusade for me, but according to [the whatwg spec](https://url.spec.whatwg.org/#goals) "URL" should be universally preferred over "URI" (contrary to Dart's unfortunate choice of class name).

lib/src/migrators/module.dart
```diff
@@ -578,17 +678,23 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+  /// could be used to depend on that stylesheet from the current one in a use,
+  /// forward, or import rule.
+  String _absoluteUriToDependency(Uri uri) {
+    var relativePath = p.relative(uri.path, from: p.dirname(_currentUrl.path));
```
```suggestion
    var relativePath = p.url.relative(uri.path, from: p.url.dirname(_currentUrl.path));
```

When you're working with URLs, you always want to use [`p.url`](https://pub.dev/documentation/path/latest/path/url.html). Otherwise you're interpreting URLs as native paths on whatever the current OS is, which will lead to all sorts of nasty edge-cases. Same throughout this function.

lib/src/migrators/module.dart
```diff
@@ -12,6 +12,7 @@ import 'package:args/args.dart';
 import 'package:sass/src/ast/sass.dart';
 
 import 'package:path/path.dart' as p;
+import 'package:sass_migrator/src/migrators/module/forward_type.dart';
```
nex3: It's generally preferred to use relative imports for files within the same `lib/`. Also below.

PR #73, opened 2019-08-02
Run Dart VM tests with snapshot

jathak:
Similar to how the Node tests require `pub run grinder js` to be run, the Dart VM tests now require `pub run grinder app-snapshot` to be run, drastically decreasing the amount of time it takes to run all of the tests.

The new `snapshot` tag is necessary for these tests now since filtering with `-t node` will still cause the check for the snapshot's existence to be run, but `-x snapshot` won't.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,11 +29,11 @@ jobs:
 
   # Testing on Node.js
   - env: NODE_VERSION=stable
-    dart_task: {test: -x snapshot}
+    dart_task: {test: -t node}
   # Keep these up-to-date with the latest LTS Node releases. They next need to
   # be rotated April 2020. See https://github.com/nodejs/Release.
   - env: NODE_VERSION=lts/dubnium
-    dart_task: {test: -x snapshot}
+    dart_task: {test: -t node}
 
   # Static checks
   - dart_task: {dartanalyzer: --fatal-warnings ./}
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,12 @@ cache:
 
 before_script:
 - if-node() { if [ ! -z "$NODE_VERSION" ]; then "$@"; fi }
+- if-dart() { if [ -z "$NODE_VERSION" ]; then "$@"; fi }
 - if-node . "$HOME/.nvm/nvm.sh"
 - if-node nvm install "$NODE_VERSION"
 - if-node nvm use "$NODE_VERSION"
 - if-node pub run grinder js
+- if-dart pub run grinder app-snapshot
 
 jobs:
   include:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -89,7 +91,7 @@ jobs:
...
+      - secure: "QCNhSiluq2TPf20ekwWlf+TkyzwnGwqS3pU+3XogRlP6ffRcG98zL9pn7cp8xmaM50x9PY6nk0t+oxkiOFw7sKhECfIRv0Z+P4cURhTUJ25+FqLPupCSqehoF93Bx8fXQR+2KCEP+cEeu2dcDU0hAZnzSWliRm/X/6Hfy+2M5eVc4x0+YC/gDjdbNbgH5P29qhRSB7klB8FtKDQdmQro0l1jSe69RGUmaeFSG7ryn+OlCJNMZqwZBd2Yb7pShiegU1ks/fIrC+pf/Yx2/VUscyLGwWqNxl6xXeZko3bNNCNPhPNIjArZJTNKMHvuIvKEvjfrUIr6TwGsw9vJmGjAbSiZzeXfeJ5zBp9ZaL1i8g5WwvVoQcbFgTn/vCisLZ97NP+0ipKr1VXxDfJ+MThaFsbOnl4wcC9XrrqQERWjH4lNGdfywvjql+d1M48veGqeYYY64UhiGDS+98t6GG+Q+ScQh5LDbn/uaQhi6DUCCqzaphdr78ylVrt7RzIdauOcu/CY3q22Kfm+wOZd+4S3Abk3L7P3eWl60sm/yUCx6yTA1rZPBylL5+R3cQS+9Lp9lll03Xcuf/L+MD7pVsN/RB5L52zn51oxtuVvjU01R0JZ8vWSgZ7CMs1BD7DXCjz9deGvBCHJR5aFelygt9ZSuIaM2xpZ2uSFscUnf6qjSh4="
```
nex3: What changed here?

.travis.yml
```diff
@@ -27,11 +29,11 @@ jobs:
 
   # Testing on Node.js
   - env: NODE_VERSION=stable
-    dart_task: {test: -t node}
+    dart_task: {test: -x snapshot}
```
nex3: Why run non-Node tests here and below?

.travis.yml
jathak: If I ignore whitespace with `git diff -w`, this goes away, so I think my editor automatically converted tabs to spaces.

.travis.yml
jathak: I forgot about the `Patch` tests and was thinking that every test was either tagged `node` or `snapshot`. Making the snapshot optional eliminated this need for this tag, so I changed it back.

jathak:
Okay, I changed it to only use the snapshot if it exists and is up-to-date and to fallback to running directly from source if not.

PR #74, opened 2019-08-13
Add flag to remove leading dashes and underscores

### LINKED ISSUE #48: Gracefully handle members that are already pseudo-private
Some libraries already prefix members with `_` or `-` to indicate that they're library-private. If any of these members are used across modules, though, leaving their names as-is will break, since they're now treated as module-private. The migrator should rename such members to remove the leading `_` or `-`.

Once https://github.com/sass/migrator/issues/47 is done, the migrator should also guarantee that members that used to start with `_` or `-` are never forwarded through top-level entrypoints (using `hide` if necessary).

jathak:
This partially resolves (but not fully) #48, allowing members that start with a dash or an underscore to be treated as library-private instead of module-private, but it's an all-or-nothing flag instead of detecting which members are used outside of the module they're declared in.

### DIFF
--- a/lib/src/migrators/module.dart
+++ b/lib/src/migrators/module.dart
@@ -44,7 +44,13 @@ class ModuleMigrator extends Migrator {
         },
         defaultsTo: 'none',
         help: 'Specifies which members from dependencies to forward from the '
-            'entrypoint.');
+            'entrypoint.')
+    ..addFlag('dash-is-library-private',
+        help: "Treats members starting with `-` and `_` as private to the "
+            "overall library, but not to the individual module. This means "
+            "that leading dashes and underscores will be removed from member "
+            "names, but these members will not be forwarded from the "
+            "entrypoint, even if --forward=all.");
 
   // Hide this until it's finished and the module system is launched.
   final hidden = true;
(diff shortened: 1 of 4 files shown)


nex3:
The more I think about this behavior, the more I don't like the way it does an unambiguously incorrect thing by default when a `_`-prefixed member is used across files (it produces `$module._name`). I think it's important for the default behavior, without any flags, to produce something at least functional if not ideal.

I think the ideal behavior here would probably be to do an initial non-modifying pass that resolves each member reference to the definition it refers to, and stores a reverse pointer as well. Then when you do a second modifying pass, you'd have the information available to determine whether there are any uses of a `_`-prefixed member outside of its module, and migrate it accordingly.

That represents a pretty big change, though, so I think it's also reasonable to just make the default be "remove all `_` prefixes" and have a flag (`--private-members`?) that opts in to preserving them. Because the user is opting into that flag, you can then throw an error if there are any cases that would end up producing something like `$module._name`.

jathak:
Yeah, that makes sense. I started adding a visitor for the first pass in a [separate branch](https://github.com/sass/migrator/blob/references/lib/src/migrators/module/references.dart) and it was less complicated than I thought it would be, so I'll finish up that when I get back from vacation and then use that for a more complete fix of #48.

The first pass visitor could probably also be useful outside of the module migrator if we wanted to make  a refactoring tool for things like member renaming (not sure if that's something that already exists for Sass)

nex3:
There's basically never been any static analysis of any kind for Sass, so you're treading new ground here :smiley:. This is the sort of thing we'll want to write a [Kythe](https://github.com/kythe/kythe) plugin as well.

PR #76, opened 2019-09-06
Track references to migrate pseudoprivate members

### LINKED ISSUE #48: Gracefully handle members that are already pseudo-private

### LINKED ISSUE #75: Update syntax for namespaced variables
`$namespace.variable` -> `namespace.$variable`

jathak:
Resolves #48.
Resolves #75.

To handle pseudoprivate members, this adds a new first pass to the module migrator that tracks member declarations and references.

This allows the main pass of the migrator to rename members when it first encounters their declarations, applying patches to every reference tracked by the first pass. This is now used for all renaming of user-defined members. When the main pass encounters member references, the only patches applied are namespaces and for built-in functions whose names change.


lib/src/migrators/module/references.dart
```diff
@@ -0,0 +1,326 @@
...
+}
+
+/// A visitor that builds a References object.
+class _ReferenceVisitor extends RecursiveAstVisitor {
+  /// The References object being built.
+  final _references = References._();
+
+  /// The current global scope.
+  ///
+  /// This persists across imports, but not across module uses.
```
nex3:
```suggestion
  /// This persists across imports, but not across module loads.
```

This better matches the terminology I've been using in docs.

lib/src/migrators/module/references.dart
```diff
@@ -0,0 +1,326 @@
...
+  ///
+  /// Note: Stylesheets only depended on through imports will not have their
+  /// own scope in this map; they will instead share a global scope with the
+  /// stylesheet that imported them.
+  final _moduleScopes = <Uri, Scope>{};
+
+  /// Namespaces present within the current stylesheet.
+  ///
+  /// Note: Unlike the similar property in _ModuleMigrationVisitor, this only
+  /// includes namespaces for use rules that already exist within the file.
```
nex3:
```suggestion
  /// includes namespaces for `@use` rules that already exist within the file.
```

It's generally good to explicitly refer to at-rules with their full names in code font, to help make it visually clear that you're not just using the corresponding English word. This is especially useful for non-native speakers.

lib/src/migrators/module/references.dart
```diff
@@ -0,0 +1,326 @@
...
+  /// It doesn't not include namespaces for to-be-migrated imports.
```
nex3:
```suggestion
  /// It doesn't include namespaces for to-be-migrated imports.
```

lib/src/migrators/module/references.dart
```diff
@@ -0,0 +1,326 @@
...
+  /// Constructs a new References object based on the stylesheet at [entrypoint]
+  /// and its dependencies.
+  static References build(Uri entrypoint) =>
+      _ReferenceVisitor().build(entrypoint);
+}
+
+/// A visitor that builds a References object.
+class _ReferenceVisitor extends RecursiveAstVisitor {
+  /// The References object being built.
+  final _references = References._();
```
nex3: Rather than constructing this and modifying it, I'd copy all its fields to the visitor and then pass them into its constructor. It's more boilerplate, but it makes the state transitions much clearer: a `References` object is now *always* an immutable bag of properties, rather than sometimes being modifiable and sometimes not.

lib/src/migrators/module/references.dart
```diff
@@ -0,0 +1,326 @@
...
+  /// This map is frozen after it is built to prevent further modification.
+  final variableReassignments = BidirectionalMap<VariableDeclaration,
+      SassNode /*VariableDeclaration|Argument*/ >();
+
+  /// Map between mixin references and their declarations.
+  ///
+  /// This map is frozen after it is built to prevent further modification.
+  final mixins = BidirectionalMap<IncludeRule, MixinRule>();
+
+  /// Map between normal function references and their declarations.
```
nex3: This doesn't include references to built-in functions, correct? Either way, it would be good to document that.

lib/src/utils.dart
```diff
@@ -80,6 +112,42 @@ Expression getOnlyArgument(ArgumentInvocation invocation) {
   }
 }
 
+/// If [node] is a get-function call whose name argument can be statically
```
```suggestion
/// If [node] is a `get-function` call whose name argument can be statically
```

lib/src/migrators/module.dart
```diff
@@ -79,17 +81,23 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
   /// Used to ensure that a dependency declaring a variable that an upstream
   /// stylesheet already declared is not treated as reassignment (since that
   /// would cause a circular dependency).
   final _upstreamStylesheets = <Uri>{};
 
+  /// Patches that rename a variable, mixin, or function.
+  ///
+  /// Unlike the main patches list, these can be generated before starting
+  /// a stylesheet's migration. This allows all patches for a declaration and
+  /// references to it to be generated at the same time.
```
nex3: What's the benefit of generating all the member rename patches at the same time?

lib/src/migrators/module.dart
```diff
@@ -113,6 +121,12 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
   /// The URL of the last stylesheet that was completely migrated.
   Uri _lastUrl;
 
+  /// A mapping between member declarations and references.
+  ///
+  /// This is generated by the first pass of the migrator and used here to
+  /// determine how a declaration seen in the second pass is used.
```
nex3: It's a little confusing to say that this is "generated by the first pass of the migrator", because the entire visitor structure is handled by `References` rather than by `_ModuleMigrationVisitor`, which might be what people think you're referring to when you say "the migrator". Maybe make it more active by saying "This performs an initial pass to determine ...".

lib/src/migrators/module.dart
```diff
@@ -174,27 +191,30 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     // they should be forwarded or not.
     var shown = <Uri, Set<String>>{};
     var hidden = <Uri, Set<String>>{};
-    categorizeMember(Uri url, String originalName, String newName) {
+    categorizeMember(
```
nex3: It's probably a good idea to document this. Even though it's local, the fact that it's complex enough to warrant factoring out suggests that it's probably also complex enough to warrant documenting.

lib/src/migrators/module.dart
```diff
@@ -284,83 +308,95 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
-          _patchRemovedColorFunction(name, arg,
-              existingArgName: _findArgNameSpan(arg));
-          name = 'adjust';
-        } else {
-          emitWarning("Could not migrate malformed '$name' call", node.span);
-          return;
-        }
+  /// If [node] is a get-function call, [getFunctionCall] should be true.
+  void _patchNamespaceForFunction(
+      FunctionExpression node, void namespacePatcher(String namespace),
```
nex3: Functions, even callbacks, should usually be verb phrases (unless they're just `callback` or `fn` or something), so I'd call `namespacePatcher` something like `patchNamespace`.

lib/src/migrators/module.dart
```diff
@@ -599,12 +621,17 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+          references.referencedOutsideDeclaringStylesheet(node)) {
+        // Remove leading `-` since private members can't be accessed outside
+        // the module they're declared in.
+        name = name.substring(1);
+      }
+      name = _unprefix(name);
       if (name != node.name) {
-        addPatch(
-            patchDelete(node.span, start: 1, end: prefixToRemove.length + 1));
+        _renameMember(node, name);
```
nex3: Style nit: make this an inline `if`. I think there are a few more you can make inline too.

lib/src/migrators/module.dart
```diff
@@ -174,27 +191,30 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
         hidden[url] ??= {};
         hidden[url].add(newName);
       }
     }
 
-    for (var entry in _scope.variables.entries) {
-      if (entry.value is! VariableDeclaration) continue;
-      categorizeMember(entry.value.span.sourceUrl,
-          (entry.value as VariableDeclaration).name, '\$${entry.key}');
+    for (var node in _scope.variables.values) {
```
nex3: I'm surprised that you still need to track the scope here. Does `References` not provide enough context?

lib/src/migrators/module.dart
jathak: Originally I thought this would simplify things by keeping all renaming in the same place and only patching namespaces when visiting references, but on second thought, this is probably more trouble than it's worth, so I've changed this back to renaming references when they are visited.

lib/src/migrators/module/references.dart
```diff
@@ -84,57 +76,38 @@ class References {
...
-    }
-    return {
-      node,
-      ...variables.keysForValue(node),
-      for (var reassignment in variableReassignments.keysForValue(node))
-        ...variables.keysForValue(reassignment)
-    };
-  }
+  References._(this.variables, this.variableReassignments, this.mixins,
+      this.functions, this.getFunctionReferences);
```
nex3: Style nit: I'd do the unmodifiable wrapping here, rather than in `_ReferenceVisitor`, just because if this *were* a public constructor that's where it would be and it's nice to have that consistency.

lib/src/util/unmodifiable_bidirectional_map_view.dart
```diff
@@ -0,0 +1,29 @@
...
+import 'bidirectional_map.dart';
+
+/// A map that allows you to efficiently find all keys associated with a
+/// particular value.
+///
+/// Each key can be associated with at most one value, but each value can be
+/// associated with more than one key.
+///
+/// This map also provides a mechanism for freezing its contents. Once [freeze]
+/// is called, this map may no longer be modified.
```
nex3: Leftover comment.

lib/src/util/unmodifiable_bidirectional_map_view.dart
```diff
@@ -0,0 +1,29 @@
...
+class UnmodifiableBidirectionalMapView<K, V> extends MapView<K, V>
```
nex3: Is this supposed to extend `UnmodifiableMapView`?

lib/src/migrators/module.dart
jathak: Okay, I think I found a way to eliminate the need for scope tracking in the main migration pass, with the exception of tracking unreferencable members from nested imports. Would it be better to move this detection to `_ReferenceVisitor` or keep tracking just these members in the main visitor? Or can you think of a way to detect these references without tracking the scope?

lib/src/migration_visitor.dart
```diff
@@ -79,7 +78,7 @@ abstract class MigrationVisitor extends RecursiveAstVisitor {
   @protected
   void visitDependency(Uri dependency, Uri source, [FileSpan context]) {
     var url = source.resolveUri(dependency);
-    var stylesheet = parseStylesheet(url);
+    var stylesheet = importCache.import(url)?.item2;
```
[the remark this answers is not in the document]
jathak: Done (I think). Let me know if there are any issues with the way I'm doing it now.

lib/src/migrators/module.dart
```diff
@@ -139,8 +136,9 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
   /// Note: We always set [migratedDependencies] to true since the module
   /// migrator needs to always run on dependencies. The `migrateFile` method of
   /// the module migrator will filter out the dependencies' migration results.
-  _ModuleMigrationVisitor(this.references, {this.prefixToRemove, this.forward})
-      : super(migrateDependencies: true);
+  _ModuleMigrationVisitor(this.importCache, this.references,
```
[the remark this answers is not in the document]

lib/src/migrators/module/references.dart
```diff
@@ -8,72 +8,64 @@
...
-  References._();
-
-  /// Map between variable references and their declarations.
+  /// An unmodifiable map between variable references and their declarations.
   ///
-  /// This map is frozen after it is built to prevent further modification.
-  final variables = BidirectionalMap<VariableExpression,
-      SassNode /*VariableDeclaration|Argument*/ >();
+  /// Each value in this map must be a [VariableDeclaration] or an [Argument].
+  final UnmodifiableBidirectionalMapView<VariableExpression,
```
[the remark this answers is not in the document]
jathak: Done

lib/src/migrators/module/references.dart
jathak: Done

lib/src/migrators/module.dart
nex3: I think I would just track this in `References`—exposed through something like `Set<SassNode> get fromNestedImports`.

lib/src/util/unmodifiable_bidirectional_map_view.dart
jathak: Yes. Fixed.

lib/src/util/unmodifiable_bidirectional_map_view.dart
jathak: Removed

lib/src/migrators/module.dart
jathak: I ended up refactoring the unreferencable-members parts of `Scope` into a separate `UnreferencableMembers` class, using the regular `Scope` in `_ReferenceVisitor` and `UnreferencableMembers` in the main visitor.

The issue with tracking it in `References` was that we need ensure both that we don't reference members from nested imports in the importing stylesheet and that we don't reference local members from the importing stylesheet within the nested import (or its dependencies). I couldn't find a way to get both without duplicating a lot of logic between the two visitors.

I also prefer that `References` represent only the current state of the stylesheets being migrated to make it potentially reusable for future migrators/static analysis, so it felt weird having it throw exceptions for unreferencable members when they don't actually become unreferencable until the nested import is migrated to `load-css`.

lib/src/migrators/module.dart
```diff
@@ -434,27 +479,32 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
-      while (current != null) {
-        _scope.variables.addAll(current.variables);
-        current = current.parent;
+    _configuredVariables = {};
+    _upstreamStylesheets.add(currentUrl);
+    if (!_useAllowed) {
+      _unreferencable = UnreferencableMembers(_unreferencable);
+      for (var declaration in references.variables.values
+          .followedBy(references.functions.values)
+          .followedBy(references.mixins.values)) {
```
nex3: Would it make sense to add `references.allDeclarations` as a convenience getter for `references.variables.values.followedBy(...).followedBy(...)`?

lib/src/migrators/module.dart
```diff
@@ -496,7 +546,8 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     var configured = <String>[];
     for (var name in locallyConfiguredVariables.keys) {
       var variable = locallyConfiguredVariables[name];
-      if (variable.isGuarded || _referencedVariables.contains(variable)) {
+      if (variable.isGuarded ||
+          references.variables.keysForValue(variable).isNotEmpty) {
```
```suggestion
          references.variables.containsValue(variable)) {
```

This seems like a generally useful method for `BidirectionalMap`.

lib/src/migrators/module/unreferencable_members.dart
```diff
@@ -0,0 +1,39 @@
...
+import 'package:sass/src/ast/sass.dart';
+
+import 'package:meta/meta.dart';
+import 'package:path/path.dart' as p;
+
+import 'package:sass_migrator/src/utils.dart';
+import 'unreferencable_type.dart';
+
+/// Tracks members that are unreferencable in the current scope.
+class UnreferencableMembers {
```
nex3: I really like making this explicitly track exactly the set of members that are unreferencable. Given that we need special handling for these (which I believe), making them super explicit keeps the code readable and the responsibilities clear.

lib/src/migrators/module/unreferencable_members.dart
```diff
@@ -0,0 +1,39 @@
...
+  final UnreferencableMembers parent;
```
nex3: Document these members.

lib/src/migrators/module/unreferencable_members.dart
```diff
@@ -0,0 +1,39 @@
...
+  final _unreferencable = <SassNode, UnreferencableType>{};
+
+  UnreferencableMembers([this.parent]);
+
+  /// Marks [declaration] as unreferencable with the given [type].
+  markUnreferencable(SassNode declaration, UnreferencableType type) {
```
nex3: Make these members explicitly `void`.

Also, the `Unreferencable` in these method names is redundant with the class name (and in practice, the instance variable name as well). I'd call this method just `add()`, to evoke a `Set` (which this is basically just a fancy version of), and I'd rename `checkUnreferencable` to just `check()`.

PR #77, opened 2019-09-16
Support loading files from load-paths and node_modules

### LINKED ISSUE #58: Gracefully handle dependencies from load paths and importers
When running in `--migrate-deps` mode, the migrator currently only ever resolves imports relative to the file that contains them. In real-world stylesheets, though, imports often come through load paths and custom importers. We need to handle these well.

For the division migrator, we can probably get away with just not traversing imports that can't be resolved, since we don't need contextual information from them to do the migration. For the module migration, it's harder; we need to know what names each import defines to be able to translate them into module references. Load paths can eventually be passed to the executable, but importers pose a bigger problem. Can we somehow have the migrator load them from JS?

See also https://github.com/sass/migrator/issues/57

### LINKED ISSUE #68: Support the webpack tilde syntax for importing from node_modules
To allow the migrator to be used for stylesheets that use the [custom webpack importer](https://github.com/webpack-contrib/sass-loader#imports) that resolves imports like

```scss
@import "~package/path/file"
```
to `node_modules/package/path/file.scss`, add support for this syntax

jathak:
Fixes #58.
Fixes #68.

This adds the `--load-path` option to the executable and also implements an approximation of the Webpack `~` syntax that resolves these URLs to stylesheets within `node_modules` directories that should be good enough to run the migrator with for most cases.

**Wait to merge until after #76**

### DIFF
--- a/lib/src/util/node_modules_importer.dart
+++ b/lib/src/util/node_modules_importer.dart
@@ -1,0 +1,60 @@
+// Copyright 2019 Google LLC
+//
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import 'package:path/path.dart' as p;
+import 'package:sass/sass.dart';
+
+import '../io.dart';
+
+/// An importer that resolves URLs starting with `~` by searching in
+/// `node_modules` directories.
+///
+/// This doesn't completely match Webpack's behavior (as it lacks the support
+/// for configuring its behavior), but it should be good enough to run the
+/// migrator with for most cases.
+class NodeModulesImporter extends Importer {
+  final _nodeModulesDirectories = <String>[];
+
+  /// Constructs a new [NodeModulesImporter] that searches for `node_modules` in
+  /// [baseDirectory] and all of its ancestors.
+  ///
+  /// If not provided, [baseDirectory] defaults to the current directory.
+  NodeModulesImporter([Directory baseDirectory]) {
+    var directory = baseDirectory ?? Directory.current;
+    while (true) {
+      var loadPath = p.join(directory.path, 'node_modules');
+      if (Directory(loadPath).existsSync()) {
+        _nodeModulesDirectories.add(loadPath);
+      }
+      var parent = p.dirname(directory.path);
+      if (directory.path == parent) break;
+      directory = Directory(parent);
+    }
+  }
+
+  /// Canonicalizes [url] using this importer.
+  ///
+  /// If [url] starts with `~`, this searches all `node_modules` directories,
+  /// using a [FilesystemImporter] to canonicalize to the real path on disk.
+  ///
+  /// If no matching stylesheet can be found, or if [url] does not start with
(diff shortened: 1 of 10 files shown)


lib/src/util/node_modules_importer.dart
```diff
@@ -0,0 +1,60 @@
...
+import '../io.dart';
+
+/// An importer that resolves URLs starting with `~` by searching in
+/// `node_modules` directories.
+///
+/// This doesn't completely match Webpack's behavior (as it lacks the support
+/// for configuring its behavior), but it should be good enough to run the
+/// migrator with for most cases.
+class NodeModulesImporter extends Importer {
+  final _nodeModulesDirectories = <String>[];
```
nex3: Nit: document this

lib/src/util/node_modules_importer.dart
```diff
@@ -0,0 +1,60 @@
...
+
+  /// Constructs a new [NodeModulesImporter] that searches for `node_modules` in
+  /// [baseDirectory] and all of its ancestors.
+  ///
+  /// If not provided, [baseDirectory] defaults to the current directory.
+  NodeModulesImporter([Directory baseDirectory]) {
```
nex3:
```suggestion
  NodeModulesImporter([String baseDirectory]) {
```

Different codebases differ on this, but I strongly prefer passing around strings for paths and only creating `File` and `Directory` objects at the last possible moment. The only state these classes have is the path anyway, and passing it as a string requires less wrapping when loading from user input and less unwrapping when passing through APIs like the `path` package.

lib/src/util/node_modules_importer.dart
```diff
@@ -0,0 +1,60 @@
...
+    var directory = baseDirectory ?? Directory.current;
```
nex3:
```suggestion
    var directory = baseDirectory ?? p.current;
```

Similar to above, and also makes it easier to switch to a custom context if you ever need to do that.

jathak:
Would something like this be acceptable?

```
<==> input/entrypoint.scss
@import "direct";

a {
  color: $variable;
}

<==> input/_direct.scss
@import "~module/indirect";

<==> input/node_modules/module/_indirect.scss
$variable: green;

<==> output/entrypoint.scss
@use "node_modules/module/indirect";
@use "direct";

a {
  color: indirect.$variable;
}

<==> output/_direct.scss
@use "~module/indirect";
```

I could change `_absoluteUrlToDependency` to add `~` URLs as one of its options, but would we want to generate those URLs even for users that weren't already using them?

nex3:
I don't love either of those options... could you track the original URLs for stylesheets and re-use them if they came from load paths or the Webpack importer?

jathak:
Okay, I think that should be doable. For detecting whether the original URL was resolved by the Webpack importer, I can just check whether the importer returned by `ImportCache.import` was the `NodeModulesImporter`, but I'm not sure how best to detect a relative import vs one from a load path.

jathak:
I'm going to merge this now. I opened #90 to use the original URLs for imports from load paths.

PR #78, opened 2019-09-18
Avoid namespace conflicts

jathak:


nex3:
WDYT about changing this to use parent directory names for these namespaces, rather than numeric suffixes? So instead of `variables` and `variables2`, you'd have `a-variables` and `b-variables`.

jathak:
Yeah, that makes sense. I'll do that.

jathak:
Hmm. How should this work for extra `@use` rules, which aren't added until we visit a reference to one of their members?

For the existing `@use` rules, we can check them all at the beginning for conflicts and then alias both conflicting rules, but for something like `@use "color"` and `@use "sass:color"`, we won't know that there's a conflict until after we've used the default namespace for the first one.

jathak:
Okay, I was able to get this working by adding additional logic to `References` to also track the source of a member being referenced (including both those from imports and those from built-in modules). The main migrator then uses that information to determine most namespaces for a stylesheet before visiting it, allowing conflicts to be resolved more intelligently. Built-in modules now get priority for their default namespaces and the migrator will now try adding additional path segments to namespaces before it falls back to numerical suffixes.

lib/src/migrators/module/references.dart
```diff
@@ -194,16 +211,57 @@ class _ReferenceVisitor extends RecursiveAstVisitor {
...
+  }
+
+  void _resolveBuiltInFunctionReferences() {
+    var functions = _unresolvedReferences.keys.whereType<FunctionExpression>();
+    for (var function in functions) {
+      if (_isCssCompatibilityOverload(function)) continue;
+      if (function.name.asPlain == null) continue;
+      var name = function.name.asPlain;
+      if (builtInFunctionModules.containsKey(name)) {
+        var module = builtInFunctionModules[name];
```
nex3: Style nit: rather than calling `containsKey()` followed by `[]`, it's usually better to just call `[]` and check if it's `null` (basically always unless the map can have meaningful `null` values). It's less repetitive and requires fewer lookups.

lib/src/migrators/module/references.dart
```diff
@@ -301,12 +381,16 @@ class _ReferenceVisitor extends RecursiveAstVisitor {
       if (!refScope.isDescendentOf(scope)) continue;
       if (reference is VariableExpression) {
         if (scope.variables.containsKey(reference.name)) {
-          _variables[reference] = scope.variables[reference.name];
+          var declaration = scope.variables[reference.name];
+          _variables[reference] = declaration;
+          _referenceSources[reference] = _declarationSources[declaration];
```
nex3: Is it worth factoring out this three-line chunk into a helper?

lib/src/migrators/module.dart
```diff
@@ -685,18 +788,45 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     return name.substring(prefixToRemove.length);
   }
 
+  /// Returns the namespace that built-in module [module] is loaded under.
+  ///
+  /// This adds an additional `@use` rule if [module] has not been loaded yet.
+  String _findBuiltInNamespace(String module) {
+    var url = builtInModuleUrls[module];
+    _namespaces.putIfAbsent(url, () => _incrementUntilAvailable(module));
```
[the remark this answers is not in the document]
jathak: Done.

lib/src/migrators/module.dart
```diff
@@ -685,18 +788,45 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+    var namespace = _namespaces[url];
+    if (!_hasUseRule.contains(url)) {
+      _hasUseRule.add(url);
+      if (namespace == module) {
```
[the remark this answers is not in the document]
jathak: Changed this to have an `asClause` variable determined by a ternary expression and then interpolated within the string literal.

lib/src/migrators/module.dart
```diff
@@ -685,18 +788,45 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+    if (!_hasUseRule.contains(url)) {
       // Add new `@use` rule for indirect dependency
-      var simplePath = _absoluteUrlToDependency(node.span.sourceUrl);
-      _additionalUseRules.add(simplePath);
-      _namespaces[node.span.sourceUrl] = namespaceForPath(simplePath);
+      var simplePath = _absoluteUrlToDependency(url);
+      var defaultNamespace = namespaceForPath(simplePath);
+      // If a member from this dependency is actually referenced, it should
+      // already have a namespace from [_determineNamespaces], so we just use
+      // a simple number suffix to resolve conflicts at this point.
```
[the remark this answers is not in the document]
jathak: There are a couple of edge cases where `references.sources` misses a source. The most common is for `@import` rules whose members aren't referenced, though that doesn't apply here since we're adding an extra `@use` rule specifically because there's a reference.

The two edge cases that I know apply here are:
- `References` doesn't track sources for functions referenced in a `get-function` call, since it's already setting that `FunctionExpression`'s source to be the built-in `meta` module. If an extra `@use` rule is only added for functions in `get-function` calls, then there won't be a namespace for that stylesheet when `_namespaceForNode` is called.
- If a variable from another module is reassigned, `References` doesn't do a good job of tracking that, so a downstream reference will think its source is the stylesheet where it's reassigned instead of where it was originally declared.

I've updated the comment here to clarify why we need to confirm that a namespace is already assigned.

lib/src/migrators/module.dart
```diff
@@ -305,22 +313,132 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
     super.visitStylesheet(node);
     _namespaces = oldNamespaces;
+    _hasUseRule = oldHasUseRule;
     _additionalUseRules = oldAdditionalUseRules;
     _lastUrl = node.span.sourceUrl;
     _useAllowed = oldUseAllowed;
   }
 
+  /// Determines namespaces for all `@use` rules that the stylesheet will
+  /// contain after migration.
```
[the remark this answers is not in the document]
nex3: Looks like `url` is still not mentioned in the doc comment.

lib/src/migrators/module.dart
```diff
@@ -318,64 +319,70 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
-        var namespace = source.defaultNamespace;
-        if (namespace != null) {
-          sourcesByNamespace.putIfAbsent(namespace, () => {}).add(source);
-        }
+    for (var reference in references.sources.keys) {
+      if (reference.span.sourceUrl != url) continue;
+      var source = references.sources[reference];
+      var namespace = source.defaultNamespace;
+      if (namespace == null) continue;
+      // Existing `@use` rules should always keep their namespaces.
```
nex3: Style nit: a newline above here would help visually differentiate the "determining which sources to add" step from the "adding those sources" step.

lib/src/migrators/module.dart
```diff
@@ -318,64 +319,70 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+      } else {
+        sourcesByNamespace.putIfAbsent(namespace, () => {}).add(source);
       }
     }
 
-    for (var namespace in sourcesByNamespace.keys) {
-      var sources = sourcesByNamespace[namespace];
-      if (sources.length == 1) {
-        _namespaces[sources.first.url] = namespace;
+    // First assign namespaces to those without conflicts.
```
```suggestion
    // First assign namespaces to module URLs without conflicts.
```

lib/src/migrators/module.dart
```diff
@@ -318,64 +319,70 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
       } else {
-        _resolveNamespaceConflict(namespace, sources);
+        conflictingNamespaces[namespace] = sources;
       }
-    }
+    });
+
+    // Then resolve conflicts where they exist.
+    conflictingNamespaces.forEach((namespace, sources) {
+      if (sources.length > 1) {
```
nex3: What happens if this conditional doesn't match? It looks like a namespace will just not be assigned, which seems wrong.

lib/src/migrators/module.dart
```diff
@@ -318,64 +319,70 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
 
-  /// Resolves a conflict between a set of sources with the same namespace.
-  void _resolveNamespaceConflict(
-      String namespace, Set<ReferenceSource> sources) {
-    // Give first priority to an existing `@use` rule, followed by a built-in
-    // module.
-    var uses = sources.where((source) => source.isUse);
-    if (uses.isNotEmpty) _namespaces[uses.first.url] = namespace;
-    var builtIns = sources.where((source) => source.isBuiltIn);
+  /// Resolves a conflict between a set of sources with the same namespace,
```
```suggestion
  /// Resolves a conflict between a set of sources with the same default namespace,
```

lib/src/migrators/module.dart
```diff
@@ -407,29 +415,46 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+        : _incrementUntilAvailable(module, existingNamespaces);
+  }
+
+  /// If [defaultNamespace] has not already been used in this
+  /// [existingNamespaces], returns it. Otherwise, returns it with the lowest
+  /// available number appended to the end.
+  ///
+  /// If [existingNamespaces] is not provided, it defaults to [_namespaces].
+  String _incrementUntilAvailable(String defaultNamespace,
+      [Map<Uri, String> existingNamespaces]) {
```
nex3: Since this is only used once without this argument being passed, it may be simpler to just make the argument mandatory and explicitly pass in `_namespaces` in that case. Same for `_resolveBuiltInNamespace`.

lib/src/migrators/module.dart
```diff
@@ -791,17 +816,14 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
   /// Returns the namespace that built-in module [module] is loaded under.
   ///
   /// This adds an additional `@use` rule if [module] has not been loaded yet.
-  String _findBuiltInNamespace(String module) {
-    var url = builtInModuleUrls[module];
-    _namespaces.putIfAbsent(url, () => _incrementUntilAvailable(module));
+  String _findBuiltInNamespaceUNSUREOFNAME(String module) {
```
```suggestion
  String _findOrAddBuiltInNamespace(String module) {
```

lib/src/migrators/module.dart
```diff
@@ -305,22 +313,132 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+      if (sources.length == 1) {
+        _namespaces[sources.first.url] = namespace;
+      } else {
+        _resolveNamespaceConflict(namespace, sources);
+      }
+    }
+  }
+
+  /// Resolves a conflict between a set of sources with the same namespace.
+  void _resolveNamespaceConflict(
```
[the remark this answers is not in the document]
nex3: SGTM

lib/src/migrators/module.dart
```diff
@@ -524,7 +618,14 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
             declaration, UnreferencableType.globalFromNestedImport);
       }
     } else {
-      _namespaces[_lastUrl] = namespaceForPath(import.url);
+      var defaultNamespace = namespaceForPath(import.url);
+      // If a member from this dependency is actually referenced, it should
+      // already have a namespace from [_determineNamespaces], so we just use
+      // a simple number suffix to resolve conflicts at this point.
```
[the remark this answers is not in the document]
nex3: Can you add a test case for this? Generally speaking, for any chunk of logic, if someone else comes along and thinks "hey this seems unnecessary what happens if I delete it?" they should see a test fail.

lib/src/migrators/module.dart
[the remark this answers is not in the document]
nex3: Same as above—please add test cases for these if they don't exist already.

lib/src/utils.dart
```diff
@@ -32,6 +32,9 @@ Patch patchAfter(AstNode node, String text) {
   return Patch(end.file.span(end.offset, end.offset), text);
 }
 
+/// Returns true if [map] does not contain any duplicate values.
+bool valuesAreUnique(Map map) => map.values.toSet().length == map.length;
```
```suggestion
bool valuesAreUnique(Map<Object, Object> map) => map.values.toSet().length == map.length;
```

It's good practice to explicitly write type arguments, even when they're just `Object`. If/when we turn on the Dart analyzer's `no-implicit-dynamic` flag, this will become mandatory.

lib/src/migrators/module.dart
jathak: Yeah, that conditional is not supposed to be present. I fixed it in one of the downstream PRs but forgot to fix it here.

lib/src/migrators/module.dart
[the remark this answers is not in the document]

lib/src/migrators/module.dart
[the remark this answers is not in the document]
jathak: Whoops. Fixed.

lib/src/migrators/module.dart

lib/src/migrators/module.dart

lib/src/migrators/module.dart

lib/src/migrators/module.dart

lib/src/utils.dart
jathak: Done

PR #91, opened 2019-09-25
Error if renaming member in unmigrated file

### LINKED ISSUE #84: --remove-prefix can generate invalid private member accesses
If you remove a prefix that's followed by a `-` in a member name, the migrator won't detect the new member as private and will accordingly generate invalid member references. For example, with `--remove-prefix=a`, `$a-b` will become `module.$-b`, which is invalid.

This seems like a likely thing for users to do by accident if they don't realize that they're supposed to include the trailing `-` in the prefix.

### LINKED ISSUE #85: --remove-prefix without --migrate-deps can generate invalid member references
If you remove a prefix without migrating dependencies, and the files you pass on the command line include the use of a prefixed member but not that member's definition, the uses will be migrated but the definition will not rendering those uses invalid. For example:

```scss
// input.scss
@import "a";
@debug $a-b;
```

```scss
// _a.scss
$a-b: c;
```

Running `sass-migrator module --remove-prefix=a- input.scss` will result in

```scss
// input.scss
@use "a";
@debug a.$b`
```

```scss
// _a.scss
$a-b: c;
```

It's probably best to just throw an error here and encourage users to re-run with either `--migrate-deps` or with `_a.scss` passed on the command line as well.

jathak:
Fixes #84.
Fixes #85.

When running the module migrator without `--migrate-deps`, it's
possible that it may generate invalid member references if it renames
a member from a dependency that is not being migrated.

This adds an error when this happens that tells the user to re-run the
migrator with `--migrate-deps` or add the dependency as an entrypoint.

### DIFF
--- a/lib/src/migrators/module.dart
+++ b/lib/src/migrators/module.dart
@@ -56,6 +56,28 @@ class ModuleMigrator extends Migrator {
         help: 'Specifies which members from dependencies to forward from the '
             'entrypoint.');
 
+  /// Set of files that declare members that the migrator wants to rename.
+  ///
+  /// If one of these files would not be migrated, the migrator will error with
+  /// a message telling the user to use `--migrate-deps` or add the missing
+  /// file to their entrypoints.
+  final _filesWithRenamedDeclarations = <Uri>{};
+
+  @override
+  Map<Uri, String> run() {
+    var results = super.run();
+    for (var file in _filesWithRenamedDeclarations) {
+      if (!results.containsKey(file)) {
+        throw MigrationException(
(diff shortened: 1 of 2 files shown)


nex3:
Is this ready to review?

lib/src/migrators/module.dart
```diff
@@ -56,6 +56,23 @@ class ModuleMigrator extends Migrator {
   // Hide this until it's finished and the module system is launched.
   final hidden = true;
 
+  final _filesWithRenamedDeclarations = <Uri>{};
```
nex3: Document this.

PR #92, opened 2019-09-25
Make all members unreferencable in nested import

### LINKED ISSUE #83: Gracefully handle imports after member definitions
Currently (on both `master` and `partial-migration`), the migrator generates invalid Sass for the following stylesheets:

```scss
// style.scss
@mixin foo { /* ... */ }
@import "other";
```

```scss
// _other.scss
@include foo;
```

It converts `style.scss` to use `meta.load-css()`, but it doesn't change `_other.scss` at all. This isn't a *super* common case, but it's something that should be handled somehow.

This is a strange case. `@include foo` references a member defined at the top level of another file, but it can't directly `@use` that file because doing so would create a circular reference. At the same time, the user probably intended to use `style.scss` as a kind of library of members usable by everything imported afterwards, which is totally reasonable. The ideal migration would probably be:

```scss
// style.scss
@use "other";
```

```scss
// _members.scss
@mixin foo { /* ... */ }
```

```scss
// _other.scss
@use "members";
@include members.foo;
```

but moving members into a totally new file would probably is likely to add a *lot* of complexity. So maybe the best thing to do right now is to add a new category of unreferencable member and throw an error encouraging the user to manually move those members into their own file.

jathak:
Fixes #83.

Previously, only local members from the importing stylesheet would be
marked as unreferencable within a nested import, but references to
global members would create a circular dependency, so they shouldn't be
allowed either.

### DIFF
--- a/lib/src/migrators/module.dart
+++ b/lib/src/migrators/module.dart
@@ -699,8 +699,7 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
       _unreferencable = UnreferencableMembers(_unreferencable);
       for (var declaration in references.allDeclarations) {
         if (declaration.sourceUrl != currentUrl) continue;
-        if (references.globalDeclarations.contains(declaration)) continue;
-        _unreferencable.add(declaration, UnreferencableType.localFromImporter);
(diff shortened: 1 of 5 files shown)


nex3:
Is this ready for review?

PR #93, opened 2019-09-26
Improve exception handling

nex3:


nex3:
This is probably a good PR to review commit-by-commit, rather than all at once.

PR #95, opened 2019-09-30
Print verbose output in HRX format

nex3:
Eventually we want to print this in diff format (see #27), but there's
not an up-to-date diff library in Dart at the moment so this is a
stop-gap measure for better output.

### DIFF
--- a/lib/src/runner.dart
+++ b/lib/src/runner.dart
@@ -106,12 +106,16 @@ class MigratorRunner extends CommandRunner<Map<Uri, String>> {
 
     if (argResults['dry-run']) {
       print('Dry run. Logging migrated files instead of overwriting...\n');
+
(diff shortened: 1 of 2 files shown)


lib/src/runner.dart
```diff
@@ -106,12 +106,16 @@ class MigratorRunner extends CommandRunner<Map<Uri, String>> {
 
     if (argResults['dry-run']) {
       print('Dry run. Logging migrated files instead of overwriting...\n');
```
jathak: Would it make sense to remove this line or make it log to stderr so that the HRX output could be piped somewhere?

nex3: I think we actually *don't* want people piping the output, since we plan to change the format to a diff at some point in the future anyway so it won't be reliably forwards-compatible.

PR #101, opened 2019-10-01
Better handling of added @use and @forward rules.

### LINKED ISSUE #46: Consistently order @uses
The module migrator currently creates `@use` rules in the order that their members are used, which ends up looking kind of messy. It would be good at some point to produce a more principled ordering. I propose:

* `sass:` `@use`s, in alphabetical order.
* Any `@use`s that come from load paths, in alphabetical order.
* Relative `@use`s, in alphabetical order.
* A newline.
* Same order for `@forward`s as above.
* A newline if there were any `@forward`s.

### LINKED ISSUE #88: Omit useless @use rules with --forward
When a `@forward` is added for a given stylesheet, that stylesheet should omit the `@use` rule for the same stylesheet unless its members are actually being used.

jathak:
Resolves #46.
Resolves #88.

`@use` and `@forward` rules should now follow the order discussed in #46, with the built-in module `@use` rules inserted immediately before the first existing import and the additional `@use` and `@forward` rules inserted immediately after the last existing import.

Code changes to make this happen:
- Instead of one `additionalUseRules` sets, this is now split into three sets for built-in modules, load paths, and relative paths.
- ~With this change, there are now eight different properties that are reset for each stylesheet in `_ModuleMigrationVisitor.visitStylesheet`. To reduce the boilerplate in this method, I extracted them out into a separate class so they can all be reset at once.~
- ~The text of a `Patch` is now mutable to allow an `@import`->`@use` patch to be changed to an `@import`->`@forward` patch without changing the patches' order (trying to track these patches separately and then add them later caused a lot of issues, mostly with `@import` rules with multiple imports)~
- I replaced `getMigratedContents` in `MigrationVisitor` with `beforePatch`, since we now needed to add patches after visiting the stylesheet but before `patches` and `currentUrl` are reset for the next stylesheet. `addPatch` now has an additional optional parameter to add patches to the beginning of the list, to resolve similar issues to the above with multiple patches at the same insertion point.


lib/src/patch.dart
```diff
@@ -11,10 +11,10 @@ class Patch {
   final FileSpan selection;
 
   /// Text to replace the selection with.
-  final String replacement;
+  String replacement;
```
nex3: I really don't like making stuff mutable if it's at all possible to avoid it. It raises so much possibility for spooky action at a distance. I'd rather see a `Patch withReplacement(String replacement)` method that returns a new `Patch` with an different replacement string.

lib/src/migrators/module.dart
```diff
@@ -325,41 +309,75 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+      } else {
+        forward += '$_semicolonIfNotIndented\n';
+        (isRelative ? relativeForwards : loadPathForwards).add(forward);
+      }
     }
-    forwards.sort();
-    return forwards.join('');
+    loadPathForwards.sort();
+    relativeForwards.sort();
+    var forwards = loadPathForwards.followedBy(relativeForwards);
```
```suggestion
    var forwards = [...loadPathForwards, ...relativeForwards];
```

It's generally a good idea to create a `List` rather than an `Iterable` if you're going to access it multiple times, since otherwise you may do lazy operations twice.

lib/src/migrators/module.dart
```diff
@@ -325,41 +309,75 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+    }
+    var extras = useRulesToString(_state.additionalLoadPathUseRules) +
+        useRulesToString(_state.additionalRelativeUseRules) +
+        _getEntrypointForwards();
+    if (extras != '') {
+      var insertionPoint = _state.afterLastImport ?? node.span.start;
+      // If there was already a blank line after the insertion point, or the
+      // insertion point was at the end of the file, remove the additional line
+      // break at the end of the extra rules.
+      var whitespace = extendThroughWhitespace(insertionPoint.pointSpan());
```
nex3: Swap this with the next line, so it's closer to where the variable is actually used.

lib/src/migrators/module.dart
```diff
@@ -667,6 +685,14 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     } else {
       addPatch(patchDelete(node.span));
     }
+    if (_state.useAllowed) {
```
nex3: Newline above this.

lib/src/migrators/module.dart
```diff
@@ -667,6 +685,14 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+      _state.beforeFirstImport ??= node.span.start;
+      var afterSemicolon =
+          extendForward(extendThroughWhitespace(node.span), ';')?.end;
+      _state.afterLastImport = currentUrl.path.endsWith('.sass')
+          ? node.span.end
+          : afterSemicolon ?? node.span.end;
```
nex3: It's a little confusing that `afterSemicolon` is computed even when it's unnecessary. I'd probably just make this an `if` statement to avoid that.

lib/src/migrators/module.dart
```diff
@@ -125,32 +126,28 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+  /// converted to `@forward` instead of `@use`.
+  ///
+  /// If an `@import` rule is visited but none of its members are referenced,
+  /// its canonical URL will be added to this, mapped to the patch that
+  /// converted it to a `@use` rule.
+  ///
+  /// When determining `@forward` rules to add based on --forward, if a URL that
+  /// would have an extra rule is a key in this map, the migrator will instead
+  /// mutate the patch to be a `@forward` rule instead of a `@use` rule.
+  final _possibleForwardConversions = <Uri, Patch>{};
```
nex3: I don't like storing patches to modify later... I hate to make such a nice self-contained class mutable, and it would be nice to keep the patch list as a write-only data structure as much as possible. To that end, would it be possible to determine in `_migrateImport()` whether a `@use` or a `@forward` should be generated, and just choose the correct one there and then?

lib/src/migrators/module.dart
```diff
@@ -782,22 +808,58 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+  /// stylesheet.
+  bool _anyMemberReferenced(Uri url) {
+    for (var entry in references.variables.entries) {
+      if (entry.key.span.sourceUrl != currentUrl) continue;
+      var declaration = _reassignedVariables.contains(entry.value)
+          ? references.variableReassignments[entry.value]
+          : entry.value;
+      if (declaration.sourceUrl == url) return true;
+    }
+    for (var entry in references.variableReassignments.entries) {
```
nex3: Separate these blocks with newlines.

lib/src/migrators/module.dart
```diff
@@ -782,22 +808,58 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+          entry.value.sourceUrl == url) {
+        return true;
+      }
+    }
+    for (var entry in references.functions.entries) {
+      if (entry.key.span.sourceUrl == currentUrl &&
+          entry.value.sourceUrl == url) {
+        return true;
+      }
+    }
```
nex3: Could you simplify this by defining `Set<SassNode> allReferences(MemberDeclaration declaration)` in `References`, and then iterating over `references.allDeclarations` rather than repeating a bunch of logic for each reference type? That helper would probably also let you simplify `References.referencedOutsideDeclaringStylesheet()`, and potentially other logic that wants to do the same thing for all members of all types.

test/migrators/module/conflicting_builtin_namespace.hrx
```diff
@@ -1,20 +0,0 @@
...
-<==> input/_color.scss
-$color: blue;
-
-<==> output/entrypoint.scss
-@use "sass:color";
-@use "color" as color2;
-
-a {
-  color: color.mix(color2.$color, red);
-}
```
nex3: Did you mean to delete this?

lib/src/migrators/module.dart
jathak: Done. I also changed this to be a method of `References`.

lib/src/migrators/module.dart

lib/src/migrators/module.dart

lib/src/migrators/module.dart

lib/src/migrators/module.dart

lib/src/migrators/module.dart
```diff
@@ -131,32 +132,28 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
-  Set<Uri> _usedUrls;
-
-  /// Set of additional `@use` rules necessary for referencing members of
-  /// implicit dependencies / built-in modules.
-  ///
-  /// This set contains the full `@use` rule without a semicolon or line break
-  /// at the end.
-  Set<String> _additionalUseRules;
+  /// Stores state for the migration of the current stylesheet.
+  StylesheetState _state;
```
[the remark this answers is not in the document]
jathak: Changed this back.

lib/src/migrators/module.dart
jathak: Yeah, I didn't think it was possible at first, but ultimately realized it was, so existing the new rule for existing `@import`s is now determined in `_migrateImport`.

lib/src/migration_visitor.dart
```diff
@@ -87,14 +87,25 @@ abstract class MigrationVisitor extends RecursiveAstVisitor {
...
     _currentUrl = oldUrl;
   }
 
+  /// Called after visiting [node], but before [currentUrl] and [patches] are
+  /// reset to their previous values.
+  ///
+  /// A migrator should override this if it needs to add any additional patches
+  /// after a stylesheet is visited.
+  @protected
+  void afterVisitingStylesheet(Stylesheet node) {}
```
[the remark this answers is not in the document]
jathak: Done

lib/src/migrators/module.dart
```diff
@@ -298,70 +300,94 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+      // `@use` rule.
+      addPatch(
+          Patch.insert(_beforeFirstImport ?? node.span.start,
+              useRulesToString(_builtInUseRules)),
+          beforeExisting: true);
+    }
+    var extras = useRulesToString(_additionalLoadPathUseRules) +
+        useRulesToString(_additionalRelativeUseRules) +
+        _getAdditionalForwardRules();
+    if (extras != '') {
```
```suggestion
    if (extras == '') return;
```

lib/src/patch.dart
```diff
@@ -48,4 +48,7 @@ class Patch {
     buffer.write(file.getText(offset));
     return buffer.toString();
   }
+
+  toString() =>
+      "'${selection.text}' (${selection.start.offset}-${selection.end.offset}) -> '$replacement'";
```
nex3: Long line

test/migrators/module/dependency_order.hrx
```diff
@@ -0,0 +1,52 @@
+<==> arguments
+--migrate-deps --forward=all --load-path=load-path
+
+<==> README
+This is designed to test the order of `@use` and `@forward` rules from multiple
+sources.
```
nex3: It's probably a good idea to describe the ordering constraints involved here.

nex3:
Forgot to mention in the code review, but this should also have a CHANGELOG entry.

jathak:
Added it in #113.

PR #102, opened 2019-10-01
Improve importing files through load paths and node_modules

nex3:
Most pertinently, we no longer migrate files in those locations, and
we properly generate @uses for dependencies within them.

### DIFF
--- a/lib/src/migration_visitor.dart
+++ b/lib/src/migration_visitor.dart
@@ -126,6 +126,9 @@ abstract class MigrationVisitor extends RecursiveAstVisitor {
       var stylesheet = result.item2;
       visitStylesheet(stylesheet);
       _importer = oldImporter;
+    } else {
+      _missingDependencies.putIfAbsent(
+          context.sourceUrl.resolveUri(dependency), () => context);
     }
   }
 
--- a/lib/src/migration_visitor.dart
+++ b/lib/src/migration_visitor.dart
@@ -116,6 +116,9 @@ abstract class MigrationVisitor extends RecursiveAstVisitor {
       // If [dependency] comes from a non-relative import, don't migrate it,
       // because it's likely to be outside the user's repository and may even be
       // authored by a different person.
+      //
+      // TODO(nweiz): Add a flag to override this behavior for load paths
+      // (#104).
       if (result.item1 != _importer) return;
 
       var oldImporter = _importer;
@@ -123,19 +126,7 @@ abstract class MigrationVisitor extends RecursiveAstVisitor {
       var stylesheet = result.item2;
       visitStylesheet(stylesheet);
       _importer = oldImporter;
-    } else {
-      handleMissingDependency(dependency, context);
     }
-  }
-
-  /// Adds the missing [dependency] within the stylesheet at [source] to the
-  /// missing dependency list to warn after migration completes.
-  ///
-  /// Migrators should override this if they want a different behavior.
-  @protected
-  void handleMissingDependency(Uri dependency, FileSpan context) {
-    _missingDependencies.putIfAbsent(
-        context.sourceUrl.resolveUri(dependency), () => context);
   }
 
   /// Returns the migrated contents of this file, or null if the file does not
--- a/lib/src/migration_visitor.dart
+++ b/lib/src/migration_visitor.dart
@@ -58,7 +58,7 @@ abstract class MigrationVisitor extends RecursiveAstVisitor {
   Uri get currentUrl => _currentUrl;
   Uri _currentUrl;
 
-  /// The importer that's currently being used to resolve relative imports.
+  /// The importer that's being used to resolve relative imports.
   ///
   /// If this is `null`, relative imports aren't supported in the current
   /// stylesheet.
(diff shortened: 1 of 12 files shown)


lib/src/migrators/module/references.dart
```diff
@@ -300,21 +311,27 @@ class _ReferenceVisitor extends RecursiveAstVisitor {
...
-                (source is CurrentSource || source is ForwardSource)) {
-              _declarationSources[declaration] = importSource;
-            }
+        if (result == null) return;
+
+        // TODO(nweiz): Rather than skipping a stylesheet we've already visited,
+        // visit it again and throw an error if it produces different results
+        // (#99).
+        var url = result.item2.span.sourceUrl;
+        if (_importers.containsKey(url)) return;
```
jathak: We need to revisit duplicate imports in `_ReferenceVisitor` (but not necessarily in `_ModuleMigrationVisitor`) to avoid breaking cases like this:

```
<==> entrypoint.scss
@import "a";
@import "b";

a {
  color: fn();
}

<==> _a.scss
@import "b";
@function fn() {@return blue;}

<==> _b.scss
@function fn() {@return green;}
```

Pre-migration, `fn()` within `entrypoint` should refer to the one in `b`, but if we only visit `b` the first time it's encountered, the definition in `a` will override it by the time the reference is visited and the main migrator will then use the wrong namespace.

lib/src/migrators/module/references.dart
```diff
@@ -361,10 +368,15 @@ class _ReferenceVisitor extends RecursiveAstVisitor {
 
   /// Given a URL from a `@use` or `@forward` rule, loads and visits the
   /// stylesheet it points to and returns its canonical URL.
-  Uri _loadUseOrForward(Uri ruleUrl) {
+  Uri _loadUseOrForward(Uri ruleUrl, AstNode nodeForSpan) {
```
jathak: Is there any reason that this needs to be `AstNode` and not `SassNode`?

lib/src/migrators/module.dart
```diff
@@ -349,14 +343,6 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     return forwards.join('');
   }
 
-  /// Immediately throw a [MigrationException] when a missing dependency is
-  /// encountered, as the module migrator needs to traverse all dependencies.
-  @override
-  void handleMissingDependency(Uri dependency, FileSpan context) {
```
jathak: Since this method only existed because the `_ModuleMigrationVisitor` needed to override it, it probably make sense to eliminate it from `MigrationVisitor`.

lib/src/migrator.dart
```diff
@@ -74,8 +75,11 @@ abstract class Migrator extends Command<Map<Uri, String>> {
   Map<Uri, String> run() {
     var allMigrated = <Uri, String>{};
     var importer = FilesystemImporter('.');
```
jathak: Is there a reason why this is not also a `ReversibleFilesystemImporter`?

lib/src/migration_visitor.dart
```diff
@@ -113,6 +113,11 @@ abstract class MigrationVisitor extends RecursiveAstVisitor {
   void visitDependency(Uri dependency, FileSpan context) {
     var result = importCache.import(dependency, _importer, _currentUrl);
     if (result != null) {
+      // If [dependency] comes from a non-relative import, don't migrate it,
+      // because it's likely to be outside the user's repository and may even be
+      // authored by a different person.
+      if (result.item1 != _importer) return;
```
jathak: This definitely makes sense for the default behavior, but would it make sense to add a flag to override this?

I think there are some stylesheets within Google that specify all dependencies relative to a load path.

nex3: Added a TODO.

lib/src/migrator.dart
nex3: URLs that are relative to the entrypoint should always be converted to relative URLs. In fact, generating a `@use` whose URL is relative to the working directory would be incorrect, since the working directory is not on the load path.

lib/src/migrators/module.dart
nex3: Done.

lib/src/migrators/module/references.dart
nex3: A good example of why it's important to have thorough test cases :smile:! Done.

lib/src/migrators/module/references.dart
nex3: Just consistency with the convention from the Sass codebase.

lib/src/migration_visitor.dart
```diff
@@ -101,24 +113,22 @@ abstract class MigrationVisitor extends RecursiveAstVisitor {
...
     }
   }
 
-  /// Adds the missing [dependency] within the stylesheet at [source] to the
-  /// missing dependency list to warn after migration completes.
-  ///
-  /// Migrators should override this if they want a different behavior.
-  @protected
-  void handleMissingDependency(Uri dependency, FileSpan context) {
-    _missingDependencies.putIfAbsent(
```
jathak: Tests are failing because this needs to be inlined into `visitDependency`

nex3: Done.

PR #110, opened 2019-10-02
Release 1.0.0

nex3:


nex3:
Yeah, it probably should. It's a bit strange because we don't have an actual release with the module system yet, but I can wait to land this until that gets published.

PR #115, opened 2019-10-10
Add parens around negated variable when namespaced

### LINKED ISSUE #114: Module Migration Does Not Handle Negating Imported Variables
**Reduced Test Case**

Given the following files:

`_variables.scss`:
```scss
$foo: 1;
```

`styles.scss`:
```scss
@import 'variables';

.foo {
    padding: -$foo;
}
```

Run the following command (sass-migrator@1.0.0):
`sass-migrator module styles.scss`

**Actual Results**

The resulting `styles.scss` throws a compilation error where the module is used.

```
@use 'variables';

.foo {
    padding: -variables.$foo;
}
```

Compilation error (sass@1.23.0):
```
> npx sass styles.scss
Error: There is no module with the namespace "-variables".
  ╷
4 │     padding: -variables.$foo;
  │              ^^^^^^^^^^^^^^^
  ╵
  styles.scss 4:14  root stylesheet
```

**Expected Results**

`styles.scss` should be migrated with no compilation errors.

Possible valid target output (this compiles successfully with sass@1.23.0):
```
@use 'variables';

.foo {
    padding: -(variables.$foo);
}
```

jathak:
Fixes #114.

In fixing this, I also found a fixed a bug when migrating removed color
functions, as a patch to namespace the argument would be applied before
the patch adding the new argument name. This was fixed by ensuring the
arguments of a function expression are visited after the function itself
is patched.

### DIFF
--- a/lib/src/migrators/module.dart
+++ b/lib/src/migrators/module.dart
@@ -522,45 +522,46 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
   /// Adds a namespace to any function call that requires it.
   @override
   void visitFunctionExpression(FunctionExpression node) {
+    if (node.namespace == null) {
+      if (references.sources.containsKey(node)) {
+        var declaration = references.functions[node];
+        _unreferencable.check(declaration, node);
+        _renameReference(nameSpan(node), declaration);
+        _patchNamespaceForFunction(node, declaration, (namespace) {
+          addPatch(patchBefore(node.name, '$namespace.'));
+        });
+      }
+
+      if (node.name.asPlain == "get-function") {
+        var declaration = references.getFunctionReferences[node];
+        _unreferencable.check(declaration, node);
+        _renameReference(getStaticNameForGetFunctionCall(node), declaration);
+
+        // Ignore get-function calls that already have a module argument.
+        var moduleArg = node.arguments.named['module'];
+        if (moduleArg == null && node.arguments.positional.length > 2) {
+          moduleArg = node.arguments.positional[2];
+        }
+        if (moduleArg != null) return;
+
+        // Warn for get-function calls without a static name.
+        var nameArg =
+            node.arguments.named['name'] ?? node.arguments.positional.first;
+        if (nameArg is! StringExpression ||
+            (nameArg as StringExpression).text.asPlain == null) {
+          emitWarning(
+              "get-function call may require \$module parameter", nameArg.span);
+          return;
+        }
+
+        _patchNamespaceForFunction(node, declaration, (namespace) {
+          var beforeParen = node.span.end.offset - 1;
+          addPatch(Patch(node.span.file.span(beforeParen, beforeParen),
+              ', \$module: "$namespace"'));
+        }, getFunctionCall: true);
+      }
+    }
     super.visitFunctionExpression(node);
-    if (node.namespace != null) return;
-    if (references.sources.containsKey(node)) {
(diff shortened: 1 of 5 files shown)


lib/src/migrators/module.dart
```diff
@@ -522,45 +522,46 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
-    if (node.namespace != null) return;
-    if (references.sources.containsKey(node)) {
-      var declaration = references.functions[node];
-      _unreferencable.check(declaration, node);
-      _renameReference(nameSpan(node), declaration);
-      _patchNamespaceForFunction(node, declaration, (namespace) {
-        addPatch(patchBefore(node.name, '$namespace.'));
-      });
-    }
+    if (node.namespace == null) {
```
nex3: Rather than fully re-indenting everything, I'd write:

```dart
if (node.namespace != null) {
  super.visitFunctionExpression(node);
  return;
}
// ...
```

I think the one duplicated line is better than the many indented lines.

lib/src/migrators/module.dart
```diff
@@ -522,45 +522,46 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+          addPatch(Patch(node.span.file.span(beforeParen, beforeParen),
+              ', \$module: "$namespace"'));
+        }, getFunctionCall: true);
       }
-
-      _patchNamespaceForFunction(node, declaration, (namespace) {
-        var beforeParen = node.span.end.offset - 1;
-        addPatch(Patch(node.span.file.span(beforeParen, beforeParen),
-            ', \$module: "$namespace"'));
-      }, getFunctionCall: true);
```
nex3: It's hard to tell for sure, but nothing actually changed in this block other than the indentation and calling `super.visitFunctionExpression()` below it instead of above it, right?

lib/src/migrators/module.dart
```diff
@@ -635,8 +636,10 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
   void _patchRemovedColorFunction(String name, Expression arg,
       {FileSpan existingArgName}) {
     var parameter = removedColorFunctions[name];
-    var needsParens =
-        parameter.endsWith('-') && arg is BinaryOperationExpression;
+    var needsParens = parameter.endsWith('-') &&
+        (arg is BinaryOperationExpression ||
+            (arg is VariableExpression &&
```
nex3: This should handle function expressions as well. Also, add a comment explaining why parentheses are needed here (just copying the one below is fine).

lib/src/migrators/module.dart
```diff
@@ -927,7 +930,12 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     _renameReference(nameSpan(node), declaration);
     var namespace = _namespaceForDeclaration(declaration);
     if (namespace != null) {
+      // Surround the variable in parens if negated to avoid `-` being parsed
+      // as part of the namespace.
+      var negated = extendBackward(node.span, '-') != null;
```
nex3: Consider writing a new `matchesBeforeSpan()` function rather than using `extendBackward()` here.

lib/src/migrators/module.dart
jathak: Yes

lib/src/migrators/module.dart

lib/src/migrators/module.dart

lib/src/migrators/module.dart

PR #122, opened 2019-10-31
Add flag to always generate import-only files

### LINKED ISSUE #121: Generate import-only files for all files with removed prefixes, not just entrypoints
I ran into this issue when trying to run material-components-web through the migrator.

Most sublibraries there consist of partials named `_functions.scss`, `_mixins.scss`, and `_variables.scss` and then an entrypoint named `mdc-<sublibrary>..scss` that, in addition to including the sublibrary's core styles, also wants to export all members from its partials.

The way to run this through the migrator is to pass the entrypoint to it with the flags `--remove-prefix=mdc-<prefix>-` and `--forward=prefixed`, which correctly forwards all of the partials through the entrypoint.

The issue is that the partials can also be depended on directly to avoid including the CSS (both by other sublibraries in material-components-web and by downstream libraries and apps), which kind of makes them "library entrypoints" from the migrator's point of view. These partials still need `.import.scss` files both to avoid breaking these downstream imports and to allow the migrator to run on other sublibraries that depend on this one.

I'm assuming this should be a flag, to avoid generating unnecessary `.import.scss` files for partials that are truly internal to a library.

@nex3: What do you think the default for this should be? (generate `.import.scss` for all files with removed prefixes or just entrypoints)

jathak:
Fixes #121.

I'm not super happy with `--import-only-for-all` as a flag name, so let me know if you think there's something simpler or clearer to use.

### DIFF
--- a/lib/src/migrators/module.dart
+++ b/lib/src/migrators/module.dart
@@ -43,6 +43,9 @@ class ModuleMigrator extends Migrator {
         abbr: 'p',
         help: 'Removes PREFIX from all migrated member names.',
         valueHelp: 'PREFIX')
+    ..addFlag('import-only-all',
+        help: 'Generates import-only stylesheets for all files with removed '
+            'prefixes, not just the entrypoint.')
     ..addOption('forward',
         allowed: ['all', 'none', 'prefixed'],
         allowedHelp: {
@@ -96,7 +99,8 @@ class ModuleMigrator extends Migrator {
         migrateDependencies: migrateDependencies,
         prefixToRemove:
             (argResults['remove-prefix'] as String)?.replaceAll('_', '-'),
-        forward: forward);
+        forward: forward,
+        importOnlyAll: argResults['import-only-all'] as bool);
     var migrated = visitor.run(stylesheet, importer);
     _filesWithRenamedDeclarations.addAll(visitor.renamedMembers.keys
         .map((declaration) => declaration.sourceUrl));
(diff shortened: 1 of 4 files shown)


lib/src/migrators/module.dart
```diff
@@ -213,29 +224,41 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
 
-  /// Generates an import-only stylesheet for [entrypoint] that forwards any
-  /// members that used to have a prefix with that prefix, but forwards other
-  /// members as-is.
+  /// Generates an import-only stylesheet for [url] that forwards any members
+  /// that used to have a prefix with that prefix, but forwards other members
+  /// as-is.
+  ///
+  /// If [url] is an entrypoint, [isEntrypoint] should be true to ensure that
+  /// any members that are forwarded through the entrypoint are also included.
```
nex3: Why don't we want that forwarding behavior for non-entrypoints? Is this difference tested?

jathak: `--forward` only forwards members from the entrypoint, not other files, so using the same behavior here for non-entrypoints would cause members from outside the file in question to be considered when determining which members need to be forwarded with a prefix and which don't.

The test I added for this flag fails if you try to use the same behavior for both entrypoints and non-entrypoints b/c this code will think `$unprefixed` is visible from `_prefix` and try to forward based on that.

PR #124, opened 2019-11-04
Support --forward=prefixed across migrator runs

### LINKED ISSUE #123: --forward=prefixed should work if the prefix was removed by a previous migrator run
For a case like:
```
// entrypoint.scss
@import "library";

// _library.scss
$variable: green;

// _library.import.scss
@forward "library" as lib-*;
```
running the migrator with `--remove-prefix=lib- --forward=prefixed` should still forward `$variable`, since it was previously prefixed within `entrypoint.scss`, but since this prefix was already removed when `_library.scss` was migrated, this isn't detected.

jathak:
Fixes #123.

If a member was unprefixed by a previous migrator run, running the
migrator on a downstream stylesheet with `--forward=prefixed` should
still forward that member.

### DIFF
--- a/lib/src/migrators/module.dart
+++ b/lib/src/migrators/module.dart
@@ -869,8 +869,8 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     var shownByPrefix = <String, Set<String>>{};
     var hidden = <String>{};
 
-    // Divide all global members from dependencies into sets based on whether
-    // they should be forwarded or not.
+    // Divide all global members from dependencies into sets based on their
+    // subprefix (if any) and whether they should be forwarded or not.
     for (var declaration in references.globalDeclarations) {
       var expectedUrl = declaration.isImportOnly
           ? declaration.forwardedUrl
@@ -893,7 +893,7 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
             importOnlyPrefix != null) {
           subprefix = importOnlyPrefix.substring(prefixToRemove.length);
         }
-        if (declaration.name.length != newName.length) _needsImportOnly = true;
+        if (declaration.name != newName) _needsImportOnly = true;
         shownByPrefix.putIfAbsent(subprefix, () => {}).add(formattedNewName);
       } else if (!newName.startsWith('-')) {
         hidden.add(formattedNewName);
--- a/lib/src/migrators/module.dart
+++ b/lib/src/migrators/module.dart
@@ -98,10 +98,8 @@ class ModuleMigrator extends Migrator {
             (argResults['remove-prefix'] as String)?.replaceAll('_', '-'),
         forward: forward);
     var migrated = visitor.run(stylesheet, importer);
-    _filesWithRenamedDeclarations.addAll({
-      for (var member in visitor.renamedMembers.keys)
-        if (member.forward == null) member.sourceUrl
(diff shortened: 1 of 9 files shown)


jathak:
Added a test with a different prefix. `--forward=prefixed` is not allowed without `--remove-prefix`, so I don't think a test is necessary for that.

jathak:
Oh, yeah, I forgot about cases like that. Would this be the best output?
```
<==> output/entrypoint.scss
@use "button";

@forward "button" as button-*;

a {
  color: button.$variable;
}

<==> output/entrypoint.import.scss
@forward "entrypoint" as lib-*;
```

nex3:
I think so, yeah.

jathak:
Okay, this should now support the subprefix case.

lib/src/migrators/module.dart
```diff
@@ -870,23 +878,52 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
-        shown.add(newName);
-      } else if (!newName.startsWith('-') && !newName.startsWith(r'$-')) {
-        hidden.add(newName);
+        var subprefix = "";
+        if (prefixToRemove != null &&
+            declaration.name.startsWith(prefixToRemove) &&
+            importOnlyPrefix != null) {
+          subprefix = importOnlyPrefix.substring(prefixToRemove.length);
+        }
+        if (declaration.name.length != newName.length) _needsImportOnly = true;
```
nex3: Why not `declaration.name != newName`?

lib/src/migrators/module.dart
```diff
@@ -870,23 +878,52 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+      if (subprefix.isNotEmpty) {
+        forward += ' as $subprefix*';
+        allHidden = [
+          for (var item in allHidden)
+            item.startsWith(r'$')
+                ? '\$$subprefix${item.substring(1)}'
+                : '$subprefix$item'
+        ];
+      }
+      if (allHidden.isNotEmpty) forward += ' hide ${allHidden.join(", ")}';
```
nex3: This generation logic is complex... are there test cases for generating multiple forwards with different `hide`s and prefixes?

lib/src/migrators/module.dart
jathak: Fixed. I was worried about an issue with underscores and dashes, but I forgot that both of these names should already be normalized to dashes.

lib/src/migrators/module.dart
```diff
@@ -853,12 +861,12 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+  /// used and a boolean that is true when the URL of these rules is relative.
   ///
   /// If nothing from [url] should be forwarded, returns null.
-  Tuple2<String, bool> _makeForwardRule(Uri url) {
-    var shown = <String>{};
+  Tuple2<List<String>, bool> _makeForwardRule(Uri url) {
+    var shownByPrefix = <String, Set<String>>{};
     var hidden = <String>{};
 
     // Divide all global members from dependencies into sets based on whether
```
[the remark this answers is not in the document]

lib/src/migrators/module.dart
jathak: Added a test for different subprefixes and fixed a bug in `References` that it revealed.

PR #126, opened 2019-11-05
Use a stable sort in Patch.applyAll

jathak:
Patches can be order-dependent when insertion patches are involved, so
we need to use a stable sort in Patch.applyAll.

### DIFF
--- a/lib/src/patch.dart
+++ b/lib/src/patch.dart
@@ -4,9 +4,10 @@
 // license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
+import 'package:collection/collection.dart';
 import 'package:source_span/source_span.dart';
 
-class Patch {
+class Patch implements Comparable<Patch> {
   /// Selection to be replaced
   final FileSpan selection;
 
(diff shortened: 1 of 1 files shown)


lib/src/patch.dart
```diff
@@ -22,8 +23,9 @@ class Patch {
 
   /// Applies a series of non-overlapping patches to the text of a file.
   static String applyAll(SourceFile file, List<Patch> patches) {
-    var sortedPatches = patches.toList()
-      ..sort((a, b) => a.selection.compareTo(b.selection));
+    var sortedPatches = patches.toList();
+    mergeSort(sortedPatches,
```
nex3: Add a comment here explaining why you're using `mergeSort()` over `List.sort()`.

lib/src/patch.dart
```diff
@@ -22,8 +23,9 @@ class Patch {
...
+        compare: (a, b) => a.selection.compareTo(b.selection));
```
nex3: I don't think you need to pass this, I think calling `compareTo()` is the default behavior.

PR #130, opened 2019-12-03
Support more complex forwards in import-only files

### LINKED ISSUE #129: Properly handle import-only files that forward other files
Right now, a case like:
```scss
// entrypoint.scss
@import "direct";

a {
  b: $direct;
  c: $indirect;
}

// _direct.scss
@use "indirect";

$direct: indirect.$indirect + 1;

// _direct.import.scss
@forward "direct";
@forward "indirect";
```

is incorrectly migrated to

```
// entrypoint.scss
@use "direct";

a {
  b: direct.$direct;
  c: direct.$indirect;
}
```

because the migrator doesn't properly account for the `@forward "indirect"` going away when migrating from `@import` to `@use`.

Fixing this is necessary for the output of #128 to be usable.

jathak:
Fixes #129.

The migrator should now correctly identify forwards that are missing when switching from an import-only file to the regular file.

This will allow libraries to require stricter dependencies for their users when those users migrate to the module system, without breaking `@import` users that depend on implicit dependencies.

To make this work, the `sourceUrl` of `MemberDeclaration` now ignores import-only files, so if a member is forwarded through an import-only file, its `sourceUrl` will be the URL will be the file being forwarded, even if it's not the regular file that corresponds to that import-only file.

This also adds `ImportOnlySource`, which is like a `ForwardSource` but for forwards within import-only files. `ImportSource` can now be created either from a real `@import` rule, or from an `ImportOnlySource`.

### DIFF
--- a/lib/src/migrators/module.dart
+++ b/lib/src/migrators/module.dart
@@ -459,7 +459,7 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
       };
       // Start each rule's namespace at the default.
       var aliases = {for (var source in sources) source: namespace};
-
+      
       // While multiple rules have the same namespace or any rule's
       // namespace is already present, add the next path segment to all
       // namespaces at once.
@@ -746,11 +746,13 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     if (migrateDependencies) visitDependency(parsedUrl, import.span);
     _upstreamStylesheets.remove(currentUrl);
 
-    // Associate the importer for this URL with the resolved URL so that we can
-    // re-use this import URL later on.
-
+    // Clear the cache for this URL and re-canonicalize it for a `@use` rule.
+    importCache.clearCanonicalize(parsedUrl);
     var tuple = inUseRule(
         () => importCache.canonicalize(parsedUrl, importer, currentUrl));
+
+    // Associate the importer for this URL with the resolved URL so that we can
+    // re-use this import URL later on.
     var resolvedUrl = tuple.item2;
     _originalImports.putIfAbsent(
         resolvedUrl, () => Tuple2(import.url, tuple.item1));
--- a/lib/src/migrators/module.dart
+++ b/lib/src/migrators/module.dart
@@ -407,11 +407,9 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     for (var reference in references.sources.keys) {
       if (reference.span.sourceUrl != url) continue;
       var source = references.sources[reference];
-      if (source is ImportSource) {
-        if (source.ruleUrl == null) {
-          source.ruleUrl =
-              _absoluteUrlToDependency(source.url, relativeTo: url).item1;
-        }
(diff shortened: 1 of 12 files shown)


lib/src/migrators/module.dart
```diff
@@ -406,6 +407,12 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     for (var reference in references.sources.keys) {
       if (reference.span.sourceUrl != url) continue;
       var source = references.sources[reference];
+      if (source is ImportSource) {
+        if (source.ruleUrl == null) {
```
Awjin: Nit: combine with parent `if`

lib/src/migrators/module/member_declaration.dart
```diff
@@ -77,3 +79,9 @@ class MemberDeclaration<T extends SassNode> {
 
   int get hashCode => member.hashCode;
 }
+
+/// Returns true if [forward] is non-null and inside an import-only file.
+bool _isImportOnly(ForwardRule forward) =>
```
Awjin: Nit: does this need to take `forward` as input?

lib/src/migrators/module/references.dart
```diff
@@ -340,21 +343,31 @@ class _ReferenceVisitor extends RecursiveAstVisitor {
...
-          if (source.url == url &&
-              (source is CurrentSource || source is ForwardSource)) {
-            _declarationSources[declaration] = importSource;
+          if (source.url == url) {
+            if (source is CurrentSource || source is ForwardSource) {
+              _declarationSources[declaration] = importSource;
+            } else if (source is ImportOnlySource) {
+              _declarationSources[declaration] =
+                  ImportSource.fromImportOnlyForward(source);
+            }
```
Awjin: IME it helps to make complex logic like this less nested for reading and debugging.

Some suggestions would be:
* on line 333, instead of having all of the logic wrapped in `if (import is DynamicImport) {}`, we could `continue` on the false condition `if (import is! DynamicImport) continue;` and eliminate a whole level of nesting.
* same on line 357, `if (source.url != url) continue;`

lib/src/migrators/module.dart
jathak: Done. I seem to recall that Dart used to not recognize type guards that were part of complex if conditions, which is why I wrote it this way, but it looks like it actually does recognize it now.

lib/src/migrators/module/member_declaration.dart
jathak: Good point. Changed this to just take a URL and moved it to `utils.dart`, so it can also be used by the references visitor.

lib/src/migrators/module/references.dart
jathak: Done. For the `DynamicImport` check, continuing if it's not one doesn't work because Dart doesn't recognize that `import` should be a `DynamicImport` for the rest of the loop, so instead I added a `whereType` condition to the for loop.

PR #131, opened 2019-12-06
Add support for --forward=import-only

### LINKED ISSUE #128: Add flag to forward all members through an import-only file (but not the regular entrypoint)
Some libraries may have downstream users that depend on members from an indirect dependency being available when they `@import` the library. Currently, libraries have to choose between forwarding all members through the regular entrypoint (which isn't ideal when ) or breaking these users when they migrate.

We should add a flag to forward members from indirect dependencies through an import-only file, but not through the regular entrypoint

jathak:
Fixes #128.

This adds a new option for `--forward`, `import-only`, which forwards
all members through the entrypoint's import-only file, but not through
the entrypoint itself.

`--forward` now supports multiple options, so a user could pass
`--forward=prefixed,import-only` to forward prefixed members through the
entrypoint _and_ forward all members through the import-only file.

This should not be merged until #130 is, as this PR is built on it and output of `--forward=import-only` only works once that PR is merged.

### DIFF
--- a/lib/src/migrators/module.dart
+++ b/lib/src/migrators/module.dart
@@ -243,8 +243,11 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     var hiddenByUrl = <Uri, Set<MemberDeclaration>>{};
     for (var declaration in references.globalDeclarations) {
       var private = declaration.name.startsWith('-');
+      // Whether this member will be exposed by the regular entrypoint.
       var visibleAtEntrypoint = declaration.sourceUrl == entrypoint ||
           (_shouldForward(declaration.name) && !private);
+      // Whether this member should be exposed by the import-only file for the
+      // entrypoint.
       var shouldBeVisible =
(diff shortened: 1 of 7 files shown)


lib/src/migrators/module.dart
```diff
@@ -217,52 +217,91 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
       var visibleAtEntrypoint = declaration.sourceUrl == entrypoint ||
-          (_shouldForward(declaration.name) &&
-              !declaration.name.startsWith('-'));
-      if (!visibleAtEntrypoint) continue;
-
-      if (declaration.name.startsWith(prefixToRemove)) {
-        forwardWithPrefix.add(declaration);
-      } else {
-        forwardWithoutPrefix.add(declaration);
+          (_shouldForward(declaration.name) && !private);
```
nex3: It's not immediately clear to me what the meaning/use of this variable is. Can you add a comment?

PR #134, opened 2020-01-24
Generate better @use rules for index files

nex3:


test/migrators/module/paths/index.hrx
```diff
@@ -0,0 +1,35 @@
...
+
+<==> input/dir4/_index.scss
+$d: 4;
+
+<==> output/entrypoint.scss
+@use "dir1";
+@use "dir2/index" as dir2;
+@use "dir3";
+@use "dir4";
+$result: dir1.$a + dir2.$b + dir3.$b + dir4.$b;
```
Awjin:
```suggestion

$result: dir1.$a + dir2.$b + dir3.$b + dir4.$b;

```

PR #136, opened 2020-02-01
Don't remove prefixes from members that would then become invalid

nex3:


lib/src/migrators/module.dart
```diff
@@ -1194,6 +1195,13 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
         isRelative);
   }
 
+  /// Returns whether [identifier] starts with [prefixToRemove], and if so,
+  /// whether the remainder is a valid Sass identifier.
+  bool _startsWithPrefix(String identifier) {
+    if (!identifier.startsWith(prefixToRemove)) return false;
+    return Parser.isIdentifier(identifier.substring(prefixToRemove.length));
+  }
+
```
nshahan: This looks unsafe to call when prefixToRemove is null and that precondition is difficult to track across all of the call sites. Is there another signal you are using to know when you can safely call _startsWithPrefix()?

nex3: Just changed this to return `false` for a null prefix, since nothing matches a prefix that doesn't exist.

PR #140, opened 2020-02-07
Fix a few bugs when migrating files that were already forwarded

nex3:


nshahan:
Is this one ready for review with the travis failure?

nex3:
Yeah, the build failures are just because we updated the `source_span` dependency.

PR #150, opened 2020-05-06
Fix pubspec issues that were preventing publishing

jathak:


pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: sass_migrator
-version: 1.2.1
+version: 1.2.1+1
```
nex3, line 2: Build identifiers are just for pre-`1.0.0` packages; this should be `1.2.2`.

pubspec.yaml
```diff
@@ -23,9 +24,9 @@ dependencies:
 
 dev_dependencies:
   archive: ">=1.0.0 <3.0.0"
+  crpyto: "^2.1.4"
```
nex3, line 27: Typo

PR #154, opened 2020-05-20
Use cli_pkg for releases

jathak:


jathak:
@nex3: Can you add `GITHUB_TOKEN` and `NPM_TOKEN` to the Travis config?

test/utils.dart
```diff
@@ -6,50 +6,36 @@
...
 import 'dart:io';
 
+import 'package:cli_pkg/testing.dart' as pkg;
 import 'package:path/path.dart' as p;
 import 'package:test/test.dart';
 import 'package:test_descriptor/test_descriptor.dart' as d;
 import 'package:test_process/test_process.dart';
 
+const jsExecutable = "build/sass-migrator.dart.js";
+const dartSnapshot = "build/sass_migrator.dart.snapshot";
```
nex3, line 16: What are these used for?

tool/grind.dart
```diff
@@ -2,13 +2,31 @@
...
+
 export 'grind/sanity_check.dart';
-export 'grind/standalone.dart';
 
-main(List<String> args) => grind(args);
+main(List<String> args) {
+  pkg.humanName = "Sass Migrator";
+  pkg.botName = "Sass Bot";
+  pkg.botEmail = "sass.bot.beep.boop@gmail.com";
+  pkg.executables = {"sass-migrator": "bin/sass_migrator.dart"};
```
nex3, line 18: This is already the autodetected value.

tool/grind.dart
```diff
@@ -2,13 +2,31 @@
...
+  pkg.homebrewRepo = "sass/homebrew-sass";
+  pkg.homebrewFormula = "migrator.rb";
+  pkg.jsRequires = {"fs": "fs", "os": "os", "path": "path"};
+  pkg.npmPackageJson =
+      json.decode(File("package/package.json").readAsStringSync())
+          as Map<String, Object>;
```
nex3, line 24: Might it make sense to just move `package.json` to the root so this is autoloaded?

tool/grind.dart
```diff
@@ -2,13 +2,31 @@
...
+  pkg.npmReadme = File("README.md").readAsStringSync();
```
nex3, line 25: This is already the autodetected value.

tool/grind.dart
```diff
@@ -2,13 +2,31 @@
...
+  pkg.standaloneName = "sass-migrator";
+
+  pkg.githubReleaseNotes = "";
```
nex3, line 28: What's up with this?

tool/grind/utils.dart
```diff
@@ -3,43 +3,25 @@
...
-
-/// The version of the migrator.
-final String version =
-    loadYaml(File('pubspec.yaml').readAsStringSync())['version'] as String;
-
-/// The version of the current Dart executable.
-final Version dartVersion = Version.parse(Platform.version.split(" ").first);
 
 /// Options for [run] that tell Git to commit using SassBot's name and email.
 final sassBotEnvironment = RunOptions(environment: {
```
nex3, line 13: Is this actually still used anywhere? Same question for a bunch of the other members in here.

.travis.yml
```diff
@@ -29,14 +29,12 @@ jobs:
 
   # Testing on Node.js
   - env: NODE_VERSION=stable
-    # Remove once pulyaevskiy/node-interop#74 is fixed.
-    dart: &node-compatible-dart
-      2.7.2
     dart_task: {test: -t node}
   # Keep these up-to-date with the latest LTS Node releases. They next need to
-  # be rotated April 2020. See https://github.com/nodejs/Release.
```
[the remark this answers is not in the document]
jathak: Will do

test/utils.dart
jathak: Whoops. These were leftover from before I fixed the bug in `cli_pkg`. Removed.

tool/grind.dart
jathak: Removed

tool/grind.dart
jathak: Done

tool/grind.dart
jathak: Removed

tool/grind.dart
jathak: Removed

tool/grind/utils.dart
jathak: I originally just modified this to remove the same members that Dart Sass did, but, yeah, it looks like the only one that was actually still used was `environment`. I just moved it and the sanity check that used it into `tool/grind.dart` and deleted the `tool/grind` directory entirely.

PR #162, opened 2020-10-12
Properly migrate orphan import-only files

### LINKED ISSUE #160: Properly support changing paths through an import-only file
Right now, the module migrator always includes a `@use` and/or `@forward` rule with the exact URL of each `@import` rule, even if that URL has an import-only file that doesn't forward the regular file.

e.g.

```
<==> entrypoint.scss
@import "old";

a { b: $lib-variable; }

<==> _old.scss
@error "Use 'new'";

<==> _old.import.scss
@forward "new" as lib-*;

<==> _new.scss
$variable: blue;
```

Right now, the entrypoint would be migrated to:
```
@use "old";
@use "new";

a { b: new.$variable }
```

While a `@use` rule with a new URL is added, and the namespacing is correct, the `@import "old";` is also migrated to a `@use`, which is wrong, since the import-only file for `old` old forwards `new`.

jathak:
Fixes #160.

### DIFF
--- a/lib/src/migrators/module.dart
+++ b/lib/src/migrators/module.dart
@@ -447,12 +447,11 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     // If there was already a blank line after the insertion point, or the
     // insertion point was at the end of the file, remove the additional line
     // break at the end of the extra rules.
-    if (insertionPoint == node.span.start) extras = '$extras\n';
+    extras = '$extras\n';
     var whitespace = extendThroughWhitespace(insertionPoint.pointSpan());
     if (whitespace.text.contains('\n\n') || whitespace.end == node.span.end) {
       extras = extras.substring(0, extras.length - 1);
     }
-    if (insertionPoint == _afterLastImport) extras = '\n$extras';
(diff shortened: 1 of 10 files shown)


lib/src/migrators/module.dart
```diff
@@ -746,7 +745,11 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     var imports =
         partitionOnType<Import, StaticImport, DynamicImport>(node.imports);
     var staticImports = imports.item1;
-    var dynamicImports = imports.item2;
+    var dynamicImports = imports.item2.where((import) =>
+        !references.orphanImportOnlyFiles.contains(importCache
+            .canonicalize(Uri.parse(import.url),
+                baseImporter: importer, forImport: true)
+            ?.item2));
```
nex3, line 752: Nit: I think this would be easier to read if you split it into separate statements.

```suggestion
    var dynamicImports = imports.item2.where((import) {
      var url = importCache.canonicalize(
              Uri.parse(import.url), baseImporter: importer, forImport: true)
          ?.item2;
      return !references.orphanImportOnlyFiles.contains(url);
    });
```

nex3 (approved):

PR #163, opened 2020-10-15
Use cli_pkg 1.0.0-beta.10

nex3:


jathak (approved):
Is there a reason why this version isn't on pub?

nex3:
It's a breaking change, but since it's in pre-release I can't give it a major version bump to ensure that it doesn't actually break existing users. To mitigate that I'm moving all the downstream users onto it first before publishing so there's no breakage period.

PR #172, opened 2021-01-30
Refactor how `@import` rules are migrated

### LINKED ISSUE #170: Issue with overlapping patches when import-only files changed paths
#168 seems to have issues with overlapping patches when two `@import`s that are both removed are adjacent to each other. I'm just inserting comments between them prior to migrating for now, but this needs to be properly fixed soon

jathak:
Fixes #170.

This refactors the migration of imports so that the entire rule is migrated in one patch, as opposed to patching each import within a rule separately. Migration for nested imports is also now factored out into its own helper (`_migrateImportToLoadCss`).

As part of this refactor, I also made three behavior changes:

- For import-only files without corresponding regular files, the regular file they actually forward will now be treated as a replacement, allowing path-changes from import-only files to be migrated in place (and avoiding the issue of adjacent imports being removed that caused #170)

- Midstream files that both forward upstream variables for downstream files to configure and configure upstream variables themselves would previously be migrated to a broken result. These cases should now be migrated to a `@forward ... with` rule, which didn't exist the last time I updated the configurable variable code.

- `@import` rules that are migrated to both a `@use` and `@forward` rule would previously only migrate one in-place (adding the extra at after all existing rules). Now, both should be migrated in place.

### DIFF
--- a/lib/src/migrators/module.dart
+++ b/lib/src/migrators/module.dart
@@ -310,7 +310,7 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
         _absoluteUrlToDependency(entrypoint, relativeTo: importOnlyUrl).item1;
     var entrypointForwards = forwardsByUrl.containsKey(entrypoint)
         ? _forwardRulesForShown(
-            entrypoint, dependency, forwardsByUrl.remove(entrypoint), {})
+            entrypoint, '"$dependency"', forwardsByUrl.remove(entrypoint), {})
         : ['@forward "$dependency"'];
     var tuples = [
       for (var entry in forwardsByUrl.entries)
@@ -322,7 +322,7 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     ];
     var forwardLines = [
       for (var tuple in tuples)
-        ..._forwardRulesForShown(tuple.item1, tuple.item2, tuple.item3,
+        ..._forwardRulesForShown(tuple.item1, '"${tuple.item2}"', tuple.item3,
             hiddenByUrl[tuple.item1] ?? {}),
       ...entrypointForwards
     ];
@@ -379,7 +379,8 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
         .map((declaration) => declaration.sourceUrl)
         .toSet()) {
       if (url == currentUrl || _forwardedUrls.contains(url)) continue;
-      var forwards = _makeForwardRules(url);
+      var forwards =
+          _makeForwardRules(url, '"${_absoluteUrlToDependency(url).item1}"');
       if (forwards == null) continue;
       var isRelative = _absoluteUrlToDependency(url).item2;
       (isRelative ? relativeForwards : loadPathForwards).addAll(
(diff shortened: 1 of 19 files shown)


lib/src/migrators/module.dart
```diff
@@ -738,110 +739,173 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+
+    var tuple = _migrateImportCommon(ruleUrl, context);
+    var canonicalUrl = tuple.item1;
+    var config = tuple.item2;
+    if (tuple.item3 != null) {
+      throw MigrationSourceSpanException(
+          "This declaration attempts to override a default value in an "
+          "indirect, nested import of ${p.prettyUri(canonicalUrl)}, which is "
+          "not possible in the module system.",
+          _configuredVariables.last.member.span);
```
nex3, line 867: It looks like this text is duplicated on line 933. I that intentional? Are there two different ways of getting this error? If so, should you add an extra test case for the other way of getting it? Should you add a shared function to throw the error so you don't have to duplicate the text?

lib/src/migrators/module/references.dart
```diff
@@ -90,9 +90,9 @@ class References {
   /// map to the [ReferenceSource] for the `sass:meta` module).
   final Map<SassNode, ReferenceSource> sources;
 
-  /// The set of import-only files that do not directly depend on their regular
-  /// counterparts.
-  final Set<Uri> orphanImportOnlyFiles;
+  /// Map of import-only files that do not directly depend on their regular
+  /// counterparts to the last forward appearing within it.
```
nex3, line 94: Mention that the value may be `null` if the file contains no `@forward`s.

lib/src/migrators/module.dart
```diff
@@ -738,110 +739,173 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+    rulesText = migratedRules.join('$_semicolonIfNotIndented\n$indent');
 
-    if (staticImports.isNotEmpty) {
-      if (dynamicImports.isNotEmpty) {
-        addPatch(Patch.insert(start, "$_semicolonIfNotIndented\n"));
-      }
+    addPatch(Patch(node.span, rulesText));
+    if (_useAllowed) {
+      _beforeFirstImport ??= node.span.start;
+      _afterLastImport = extendThroughLine(node.span).end;
```
nex3, line 786: What happens here if the user's entire stylesheet is on one line?

nex3 (approved):

lib/src/migrators/module.dart
jathak: Yeah, there's only supposed to be one. I factored this out into `_migrateImportToLoadCss` but forgot to remove it from the common logic. Fixed.

lib/src/migrators/module/references.dart
jathak: Done.

lib/src/migrators/module.dart
jathak: Good catch. Modified this to add proper logic for finding the safe place to insert additional rules (ideally after a line break, but will leave this insertion point in the middle of a line if the file ends or if there's additional text that's not a comment or whitespace on the same line).

lib/src/utils.dart
```diff
@@ -76,6 +76,47 @@ FileSpan extendForward(FileSpan span, String text) {
   return span.file.span(span.start.offset, end + text.length);
 }
 
+/// Returns the next location after [import] that it would be safe to insert
+/// a `@use` or `@forward` rule.
+///
+/// This is generally the start of the next line, but may vary if there's any
+/// non-whitespace, non-comment code on the same line.
+FileLocation afterImport(ImportRule import, {bool shouldHaveSemicolon = true}) {
```
nex3, line 84: You might be able to simplify this by comparing the end location of `import` with the start location of the next statement in `import.parent.children` (ignoring comments). If they're on different lines, you can just choose the end-of-line location; if they're not, you can choose the beginning of the next statement instead.

jathak: I considered that, but `SassNode`s don't actually have a `parent` property (only `ModifiableCssNode`s do as far as I can tell).

PR #173, opened 2021-02-04
Adds the namespace migrator

jathak:
This allows you to change namespaces for `@use` rules by matching regular expressions on the existing namespace or on the rule URL.

The migrator also has special handling for `@use` rules without member references which can be configured to run only when they conflict with a referenced `@use` rule (default), for all unreferenced rules, or for none of them.

The renaming itself is specified through a small DSL. The basics of this are documented in the changelog and the help text, but I'll add proper documentation to the website before merging this and releasing 1.3.0.

The tests under `test/migrators/namespace` focus on conflict resolution and the actual patching of source files, while the DSL itself is mostly tested in the unit tests in `test/renamer_test.dart`.


lib/src/migrators/namespace.dart
```diff
@@ -0,0 +1,245 @@
...
+      ImportCache importCache, Stylesheet stylesheet, Importer importer) {
+    Renamer<UseRule> renamer;
+    try {
+      renamer = Renamer(argResults['rename'].join('\n'), 'namespace', {
+        'namespace': (rule) => rule.namespace,
+        'url': (rule) => rule.url.toString()
+      });
+    } on FormatException catch (e) {
+      throw MigrationException(e.message);
+    }
```
nex3, line 68: Also catch a `SourceSpanFormatException` and convert it to a `MigrationSourceSpanException`.

lib/src/migrators/namespace.dart
```diff
@@ -0,0 +1,245 @@
...
+    var oldUsedNamespaces = _usedNamespaces;
+    _spansByNamespace = {};
+    _usedNamespaces = {};
+    super.visitStylesheet(node);
+    _spansByNamespace = oldSpansByNamespace;
+    _usedNamespaces = oldUsedNamespaces;
+  }
+
+  @override
+  void beforePatch(Stylesheet node) {
```
nex3, line 105: Breaking this up with some empty lines between blocks would help readability.

lib/src/migrators/namespace.dart
```diff
@@ -0,0 +1,245 @@
...
+    // Pass each `@use` rule through
```
nex3, line 106: Through what?

lib/src/migrators/namespace.dart
```diff
@@ -0,0 +1,245 @@
...
+  }
+}
+
+class _NamespaceMigrationVisitor extends MigrationVisitor {
+  final Renamer<UseRule> renamer;
+  final UnreferencedFlag unreferencedFlag;
+  final bool forceRename;
+
+  Map<String, Set<FileSpan>> _spansByNamespace;
+  Set<String> _usedNamespaces;
```
nex3, line 87: Document these.

lib/src/migrators/namespace.dart
```diff
@@ -0,0 +1,245 @@
...
+          if (!_spansByNamespace.containsKey(rule.namespace)) {
+            unreferencedRules.add(rule);
+            rules.remove(rule);
+          }
+        }
+      }
+      if (rules.length == 1) {
+        _patchNamespace(rules.first, newNamespace);
+        continue;
+      }
```
nex3, line 138: Dead code?

lib/src/migrators/namespace.dart
```diff
@@ -0,0 +1,245 @@
...
+      if (!forceRename) {
+        throw MigrationException(
+            'Rename failed. ${rules.length} rules would use namespace '
+            '"$newNamespace".\n'
+            'Run with --force to rename with numerical suffixes.');
```
nex3, line 143: This should include the spans of the rules in question. You can hook into `MultiSpanSassException` to make that easy.

lib/src/migrators/namespace.dart
```diff
@@ -0,0 +1,245 @@
...
+      _patchNamespace(rule, namespace);
+    }
+  }
+
+  /// Patch [rule] and all references to it with [newNamespace].
+  void _patchNamespace(UseRule rule, String newNamespace) {
+    _usedNamespaces.add(newNamespace);
+    if (rule.namespace == newNamespace) return;
+    var asClause =
+        RegExp('\\sas\\s(${rule.namespace})').firstMatch(rule.span.text);
```
nex3, line 171: It's possible that there's multiple characters of whitespace between `as` and the namespace. It's even vaguely possible that there's no whitespace before `as`, although I'm not sure we really need to worry about that.

If it's helpful, you could add `UseRule.urlSpan`, which would make this and the replacement logic below much more reliable.

lib/src/migrators/namespace.dart
```diff
@@ -0,0 +1,245 @@
...
+    if (asClause == null) {
+      // Add an `as` clause to a rule that previously lacked one.
+      var end = RegExp(r"""@use\s("|').*\1""").firstMatch(rule.span.text).end;
```
nex3, line 174:
```suggestion
      var end = RegExp(r"""@use\s("|').*?\1""").firstMatch(rule.span.text).end;
```

If the `.*` here is greedy, it would match all the way into a `with` clause that contained quoted string arguments, which is definitely not what we want.

test/renamer_test.dart
```diff
@@ -0,0 +1,174 @@
...
+      });
+    });
+
+    group('escaping', () {
+      test(r'spaces', () {
+        var renamer = Renamer.simple(r'a\ b to x\ y');
+        expect(renamer.rename('a b'), equals('x y'));
+      });
+
+      test(r'normal delimiters', () {
```
nex3, line 84: Also test a delimiter at the beginning of the source pattern and escaped backslashes, including things like `foo to bar\\;` and `foo to bar\\\;`.

test/renamer_test.dart
```diff
@@ -0,0 +1,174 @@
...
+            Renamer.map(r'url .*/(\w+)/lib/mixins to \1', ['namespace', 'url']);
+        expect(
+            renamer.rename(
+                {'namespace': 'mixins', 'url': 'path/button/lib/mixins'}),
+            equals('button'));
+      });
+    });
+  });
+
+  group('multiple statements', () {
```
nex3, line 116: Test empty initial statements. In particular, it's likely that users writing shell scripts will write empty initial newlines.

test/renamer_test.dart
```diff
@@ -0,0 +1,174 @@
...
+
+    test('only first matching statement is applied', () {
+      var renamer = Renamer.simple('/.*/all/;/old/wrong/;/all/wrong/');
+      expect(renamer.rename('old'), equals('all'));
+      expect(renamer.rename('wrong'), equals('all'));
+      expect(renamer.rename('all'), equals('all'));
+    });
+  });
+
+  group('invalid syntax', () {
```
nex3, line 149: Also:

* Empty string
* Three clauses but the middle one isn't `to`
* Four clauses but the first one isn't a known key

test/utils.dart
```diff
@@ -132,7 +132,8 @@ class _HrxTestFiles {
             "given test.";
       }
     } else if (filename == "arguments") {
-      arguments = contents.trim().split(" ");
+      // TODO(jathak): Handle spaces with quotes here.
```
nex3, line 135: Not for the first time, I really wish Dart had a shellparse package. In this case, I think you should be able to make do with scanning for instances of `"[^"]+"|[^"\s][^\s]*`

nex3 (changes requested):

test/renamer_test.dart
jathak: Done. The updated parser should handle any amount of spaces, newlines, and semicolons at any point before or after a statement.

test/renamer_test.dart
jathak: Done. I decided to make a renamer with no statements not actually an error (it just will always return null), so that if someone wants to run this migrator just for `--unreferencable=all` it should still work.

test/utils.dart
jathak: Done.

jathak (commented):
Decided to remove the sed-style syntax after finding some edge cases where it's ambiguous which syntax a statement should be parsed as.

I rewrote the parser accordingly using StringScanner.

lib/src/renamer.dart
```diff
@@ -0,0 +1,259 @@
...
+
+  /// Reads the next statement (and the trailing delimiter, if any) from
+  /// [scanner].
+  static _Statement<T> _readStatement<T>(
+      StringScanner scanner, Map<String, String Function(T input)> keys) {
+    var start = scanner.position;
+    FormatException lastException;
+    // Tries each key in succession until one is successfully returned.
+    for (var entry in keys.entries) {
+      try {
```
nex3, line 100: If I'm understanding correctly, the reason you have a `try`/`catch` here is to handle the case where a user uses the default key is used for a pattern that's also the same as a key name (e.g. `url to my-url`), right? That's a complex enough case that it's worth documenting in a comment (and testing if you aren't already).

For reference, catching exceptions like this to handle control flow is quite slow. This code is not performance-critical at all so it probably doesn't matter, but if you wanted to make something like this more efficient you could potentially do so by doing a "lexing" pass where you handle whitespace separators and escape sequences and return a list of strings in the next statement, and then a "parsing" phase where you can make decisions based on how many strings there are.

lib/src/renamer.dart
```diff
@@ -0,0 +1,259 @@
...
+        scanner.position = start;
```
nex3, line 101: I'd do this in the `catch` clause, so that it's clear that the reason the position is in a weird state is because you're handling an exception.

lib/src/renamer.dart
```diff
@@ -0,0 +1,259 @@
...
+    var output = _readOutput(scanner);
+    return _Statement(keyFunction, matcher, output);
+  }
+
+  /// Reads a matcher clause and its trailing space from [scanner].
+  static RegExp _readMatcher(StringScanner scanner) {
+    int quote;
+    if ({$single_quote, $double_quote}.contains(scanner.peekChar())) {
+      quote = scanner.readChar();
+    }
```
nex3, line 135: I think people who are used to JS or shell idioms would get pretty confused using quotes to delimit a regular expression. I'd just stick with allowing spaces to be backslash-escaped, especially since in the use-cases we know of we expect it to be very rare that someone would want to write a space in the input or output. As a bonus, that'll also make this logic simpler :smile:.

nex3 (changes requested):
Looking good!

CHANGELOG.md
```diff
@@ -1,3 +1,43 @@
...
+  By default, if the renaming results in a conflict between multiple `@use`
+  rules, the migration will fail, but you can force it to resolve conflicts with
+  numerical suffixes by passing `--force`.
+
+  This migrator will also check for `@use` rules without referenced members. If
+  any of these results has a conflict, it will be resolved by namespacing the
+  unreferenced rule with `_unreferenced#` instead of failing. You can disable
+  this behavior with `--unreferenced=none` or have the migrator rename *all*
+  unreferenced `@use` rules with this pattern (instead of just conflicting ones)
+  by passing `--unreferenced=all`.
```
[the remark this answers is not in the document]
jathak: Removed

lib/src/renamer.dart
```diff
@@ -0,0 +1,259 @@
...
+    return Renamer(code, {for (var key in keys) key: (input) => input[key]});
+  }
+
+  /// Creates a Renamer based on [code] and a map from keys to functions that
+  /// take an input and return a string.
+  ///
+  /// All keys must consist of only lowercase letters, underscores, and hyphens.
+  ///
+  /// If provided, [sourceUrl] will appear in parsing errors. It can be
+  /// a [String] or a [Uri].
```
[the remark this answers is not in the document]
jathak: Added. This was necessary to provide some useful context to any exceptions that StringScanner throws (I pass `--rename` as the source "URL" so that any parsing exceptions there indicate that it's from the rename option, and not from an actual Sass file.

lib/src/renamer.dart
jathak: Added a comment and a couple of tests.

I considered that kind of two-pass parse, but the matcher and output clauses have slightly different behaviors for escape sequences (since the matcher clause needs to leave most of them intact for the regex parser to handle, but the output clause handles them all). Because of that, I figured the performance cost of the exceptions was worth the simpler implementation given that only the parsing of the renamer scripts would be slowed, not the actual execution.

lib/src/renamer.dart
jathak: Done

lib/src/renamer.dart
jathak: Removed

nex3 (approved):

PR #174, opened 2021-02-09
`--remove-prefix` fixes for private members

### LINKED ISSUE #169: Private members unnecessarily hidden by @forward rules in import-only files

jathak:
Fixes #169.

Unnecessary hide clauses will no longer be added to import-only files.
Prefixes can now also be removed from private members.

### DIFF
--- a/lib/src/migrators/module.dart
+++ b/lib/src/migrators/module.dart
@@ -245,9 +245,11 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     var hiddenByUrl = <Uri, Set<MemberDeclaration>>{};
     for (var declaration in references.globalDeclarations) {
       var private = declaration.name.startsWith('-');
+
       // Whether this member will be exposed by the regular entrypoint.
-      var visibleAtEntrypoint = declaration.sourceUrl == entrypoint ||
-          (_shouldForward(declaration.name) && !private);
+      var visibleAtEntrypoint = (declaration.sourceUrl == entrypoint ||
+              _shouldForward(declaration.name)) &&
+          !private;
       // Whether this member should be exposed by the import-only file for the
       // entrypoint.
       var shouldBeVisible =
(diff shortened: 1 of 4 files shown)


lib/src/migrators/module.dart
```diff
@@ -245,9 +245,11 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     var hiddenByUrl = <Uri, Set<MemberDeclaration>>{};
     for (var declaration in references.globalDeclarations) {
       var private = declaration.name.startsWith('-');
+
       // Whether this member will be exposed by the regular entrypoint.
-      var visibleAtEntrypoint = declaration.sourceUrl == entrypoint ||
-          (_shouldForward(declaration.name) && !private);
+      var visibleAtEntrypoint = (declaration.sourceUrl == entrypoint ||
+              _shouldForward(declaration.name)) &&
+          !private;
```
nex3, line 252:
```suggestion
      var visibleAtEntrypoint = !private &&
          (declaration.sourceUrl == entrypoint || _shouldForward(declaration.name));
```

Nit: I think this would read a bit better (and possibly be a bit more efficient).

lib/src/migrators/module.dart
```diff
@@ -1167,8 +1173,11 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
   ///
   /// Otherwise, returns [name] unaltered.
   String _unprefix(String name) {
-    var prefix = _prefixFor(name);
-    return prefix == null ? name : name.substring(prefix.length);
+    var unprivateName = name.startsWith('-') ? name.substring(1) : name;
+    var prefix = _prefixFor(unprivateName);
+    if (prefix == null) return name;
+    return (name == unprivateName ? '' : '-') +
```
nex3, line 1179: Rather than `name == unprivateName` here, consider adding an `isPrivate` variable above.

nex3 (approved):

PR #175, opened 2021-03-10
Migrate CI to GitHub Actions

jathak:
We weren't previously testing on Windows, so this caught a bug with load paths that is also fixed in this PR


tool/grind.dart
```diff
@@ -23,10 +23,9 @@ main(List<String> args) {
 
 @Task('Verify that the package is in a good state to release.')
 sanityCheckBeforeRelease() {
-  var travisTag = environment("TRAVIS_TAG");
-  if (travisTag != pkg.version.toString()) {
-    fail("TRAVIS_TAG $travisTag is different than pubspec version "
-        "${pkg.version}.");
+  var ref = environment("GITHUB_REF");
+  if (ref != "refs/tags/${pkg.version}") {
+    fail("TRAVIS_TAG $ref is different than pubspec version ${pkg.version}.");
```
nex3, line 28:
```suggestion
    fail("GITHUB_REF $ref is different than pubspec version ${pkg.version}.");
```

nex3 (approved):
I'm guessing I'll need to add the various secrets to GH actions for this repo in order for it to be deployable?

jathak:
> I'm guessing I'll need to add the various secrets to GH actions for this repo in order for it to be deployable?

Yes. Does that need to be done before merging this or after?

nex3:
After is fine, it just needs to happen before the next deploy.

PR #177, opened 2021-03-22
Fix a crash with orphan import-only files

jathak:
Previously we just used the rule URL from the orphan import-only's last
`@forward` rule directly, but that only works when that URL is relative
to a load path or the orphan is in the same directory as the file
depending on it.

We now build a new dependency URL after resolving the old rule URL into
its canonical form.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 1.3.4
+
+### Module Migrator
+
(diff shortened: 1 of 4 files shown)


jathak:
@Awjin Can you review this? (GitHub won't let me assign you as a reviewer)

Awjin:
LGTM

PR #180, opened 2021-03-30
Fix one more orphan import-only bug

jathak:


jathak:
@Awjin: At this point, I feel like I'm cursed to find a bug with orphan import-only files that only manifests under very specific conditions in google3 every week or so. PTAL when you get the chance.

Awjin:
LGTM

PR #183, opened 2021-04-24
Migrate to null safety

jathak:


jathak:
Windows tests are failing due to an issue cloning a git dependency

lib/src/migrators/namespace.dart
```diff
@@ -83,7 +83,7 @@ class _NamespaceMigrationVisitor extends MigrationVisitor {
   @override
   void beforePatch(Stylesheet node) {
     // Pass each `@use` rule through the renamer.
-    var newNamespaces = <String, Set<UseRule>>{};
+    Map<String?, Set<UseRule>> newNamespaces = <String, Set<UseRule>>{};
```
nex3, line 86: This looks like a bogus retype from the migrator. I tink the `rule.namespace == null` check below should be made into an intemediate variable, at which point this can genuinely have non-nullable keys.

lib/src/migrators/namespace.dart
```diff
@@ -60,10 +60,10 @@ class _NamespaceMigrationVisitor extends MigrationVisitor {
   /// A set of spans for each *original* namespace in the current file.
   ///
   /// Each span covers just the namespace of a member reference.
-  Map<String, Set<FileSpan>> _spansByNamespace;
+  Map<String, Set<FileSpan>>? _spansByNamespace;
 
   /// The set of namespaces used in the current file *after* renaming.
-  Set<String> _usedNamespaces;
+  Set<String>? _usedNamespaces;
```
nex3, line 66: Looks like these should have non-null assertion wrappers.

lib/src/migrators/namespace.dart
```diff
@@ -187,7 +188,7 @@ class _NamespaceMigrationVisitor extends MigrationVisitor {
   void visitIncludeRule(IncludeRule node) {
     if (node.namespace != null) {
       var startNamespace = node.span.text.indexOf(
-          node.namespace, node.span.text[0] == '+' ? 1 : '@include'.length);
+          node.namespace!, node.span.text[0] == '+' ? 1 : '@include'.length);
```
nex3, line 191: Intermediate variable

lib/src/renamer.dart
```diff
@@ -159,7 +166,7 @@ class Renamer<T> {
           src.writeCharCode(next);
         }
       } else {
-        src.writeCharCode(char);
+        src.writeCharCode(char!);
```
nex3, line 169: `char` could actually be null here if the string ended unexpectedly.

lib/src/renamer.dart
```diff
@@ -183,7 +190,7 @@ class Renamer<T> {
           buffer.writeCharCode(next);
         }
       } else {
-        buffer.writeCharCode(char);
+        buffer.writeCharCode(char!);
```
nex3, line 193: If you make the `if ({...}.contains(char))` check above more explicit you won't need this assertion.

lib/src/util/bidirectional_map.dart
```diff
@@ -43,16 +43,17 @@ class BidirectionalMap<K, V> extends MapBase<K, V> {
...
+    var keys = _keysForValue[value]!;
+    keys.remove(key);
+    if (keys.isEmpty) _keysForValue.remove(value);
     return value;
   }
 
   /// Finds the keys associated with a given value.
   Iterable<K> keysForValue(V value) sync* {
-    if (_keysForValue.containsKey(value)) yield* _keysForValue[value];
+    if (_keysForValue.containsKey(value)) yield* _keysForValue[value]!;
```
nex3, line 57: Intermediate variable

lib/src/utils.dart
```diff
@@ -135,11 +138,11 @@ bool isWhitespace(int character) =>
 /// `$` at the start of variable names.
 FileSpan nameSpan(SassNode node) {
   if (node is VariableDeclaration) {
-    var start = node.namespace == null ? 1 : node.namespace.length + 2;
+    var start = node.namespace == null ? 1 : node.namespace!.length + 2;
```
nex3, line 141: Intermediate variable, also below

test/utils.dart
```diff
@@ -63,7 +65,7 @@ Future<void> _testHrx(File hrxFile, String migrator) async {
 
   if (files.expectedLog != null) {
     expect(process.stdout,
-        emitsInOrder(files.expectedLog.trimRight().split("\n")));
+        emitsInOrder(files.expectedLog!.trimRight().split("\n")));
```
nex3, line 68: Intermediate variable

tool/grind.dart
```diff
@@ -43,7 +43,7 @@ sanityCheckBeforeRelease() {
 /// Returns the environment variable named [name], or throws an exception if it
 /// can't be found.
 String environment(String name) {
-  var value = Platform.environment[name];
+  String? value = Platform.environment[name];
```
nex3, line 46: I think this type annotation is unnecessary

nex3 (changes requested):

lib/src/utils.dart
```diff
@@ -17,6 +17,9 @@ import 'package:sass/src/ast/node.dart';
 import 'io.dart';
 import 'patch.dart';
 
+T assertNotNull<T>(T? object) =>
+    object ?? (throw AssertionError('Unexpected null value.'));
```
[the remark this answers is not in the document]
jathak: Done.

All but one use of this was for a variable that's initialized in `visitStylesheet` of a `MigrationVisitor`, so I added a method there. `_configuredVariables` is initialized when a dependency is visited, but since it's the only case like that I just threw the error inline.

lib/src/migrators/module.dart
```diff
@@ -1290,12 +1307,12 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
   ///
   /// If there is no such prefix, returns `null`.
-  String _prefixFor(String identifier) => maxBy(
+  String? _prefixFor(String identifier) => maxBy(
       prefixesToRemove.where((prefix) =>
           prefix.length < identifier.length &&
           identifier.startsWith(prefix) &&
           Parser.isIdentifier(identifier.substring(prefix.length))),
-      (prefix) => prefix.length);
+      (prefix) => prefix!.length);
```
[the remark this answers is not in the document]
nex3: It's probably worth filing this as a bug against the `collection` package.

lib/src/migrators/module.dart
```diff
@@ -764,17 +777,17 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
     for (var import in dynamicImports) {
-      var ruleUrl = import.url;
+      String? ruleUrl = import.url;
       var tuple = importCache.canonicalize(Uri.parse(ruleUrl),
           baseImporter: importer, baseUrl: currentUrl, forImport: true);
       var canonicalImport = tuple?.item2;
       if (references.orphanImportOnlyFiles.containsKey(canonicalImport)) {
         ruleUrl = null;
-        var url = references.orphanImportOnlyFiles[canonicalImport]?.url;
+        var url = references.orphanImportOnlyFiles[canonicalImport!]?.url;
```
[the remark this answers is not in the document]
nex3: I meant `canonicalImport`, not the result of `Map.[]`.

lib/src/migrators/module.dart
```diff
@@ -1219,17 +1248,16 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
     var url = declaration.sourceUrl;
     if (url == currentUrl) return null;
 
     // If we can load [declaration] from a library entrypoint URL, do so. Choose
     // the shortest one if there are multiple options.
     var libraryUrls = references.libraries[declaration];
-    if (libraryUrls != null) {
-      url = minBy(libraryUrls, (url) => url.pathSegments.length);
+    if (libraryUrls != null && libraryUrls.isNotEmpty) {
+      Uri? minUrl = minBy(libraryUrls, (url) => url.pathSegments.length);
```
nex3, line 1259: Generally I think it's cleaner to try to keep all type stuff on the RHS if possible, so if inference isn't working here I'd write:

```suggestion
      var minUrl = minBy<Uri>(libraryUrls, (url) => url.pathSegments.length);
```

Same below.

nex3 (approved):

lib/src/migrators/module.dart
jathak: `minBy` has two type parameters, so is this still cleaner as `minBy<Uri, int>`?

nex3: Yeah, I think so.

PR #186, opened 2021-05-12
Update the division migrator to use modules

jathak:
This also prepares the division migrator for release in 1.4.0.

### DIFF
--- a/lib/src/migrators/division.dart
+++ b/lib/src/migrators/division.dart
@@ -23,13 +23,9 @@ import 'package:sass_migrator/src/utils.dart';
 class DivisionMigrator extends Migrator {
   final name = "division";
   final description = """
-Use the divide() function instead of the / division operator
+Use the math.div() function instead of the / division operator
 
 More info: https://sass-lang.com/d/slash-div""";
-
-  // Hide this until the division deprecation has been launched in at least one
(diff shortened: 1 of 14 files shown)


lib/src/migrators/division.dart
```diff
@@ -64,6 +60,74 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
...
+  /// will be an empty string if [module] is already used with no namespace.
+  ///
+  /// If [module] is not already used in this file, a new `@use` rule will be
+  /// added to [_useRulesToInsert].
+  String _builtInPrefix(String module) {
+    var url = Uri.parse('sass:$module');
+    if (_existingNamespaces.containsKey(url)) {
+      return _existingNamespaces[url].andThen((ns) => '$ns.') ?? '';
+    }
+    options() sync* {
```
nex3, line 114: Add a return type, otherwise you'll get `dynamic`s everywhere.

nex3 (approved):

lib/src/migrators/division.dart
jathak: The analyzer does seem to pick up that this is an `Iterable<String>` without the explicit type, but I'll add one just to be safe

PR #193, opened 2021-05-26
Properly migrate negated parenthesized expressions

### LINKED ISSUE #192: Surrounding minus operator is not taken into account when migrating division operator
Not sure if this should be a migration issue or a dart-sass issue, I would like dart-sass to change this behavior and allow minus operator to be present before a function invocation, but _meanwhile_, this tool presents an issue.

Before migration:
```scss
@use 'sass:math';

body {
  font-size: -(1/2);
}
```

After migration:
```scss
@use 'sass:math';

body {
  font-size: -math.div(1, 2);
}
```

Error:
```scss
echo "@use 'sass:math'; body {font-size: -math.div(1, 2)}" | npx sass --stdin

Error: There is no module with the namespace "-math".
  ╷
1 │ @use 'sass:math'; body {font-size: -math.div(1, 2)}
  │                                    ^^^^^^^^^^^^^^^
  ╵
  - 1:36  root stylesheet
```

Expected migration:
```scss
@use 'sass:math';

body {
  font-size: -(math.div(1, 2));
}
```

Expected result:
```scss
body {
  font-size: -0.5;
}
```

Additionally I'm aware you can just use -1 / 2, but this issue is about surrounding the content using minus operator. I see a few of those cases during migration and it's just inconvenient to have to go fix those. Whether this issue is within the scope of the migration tool is up to you of course.

jathak:
Fixes #192.

### DIFF
--- a/lib/src/migrators/division.dart
+++ b/lib/src/migrators/division.dart
@@ -60,6 +60,9 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
   /// True when the current node is expected to evaluate to a number.
   var _expectsNumericResult = false;
 
+  /// True when visiting a negated [ParenthesizedExpression].
+  var _negatedParenthesized = false;
+
   /// The namespaces that already exist in the current stylesheet.
   Map<Uri, String?> get _existingNamespaces =>
       assertInStylesheet(__existingNamespaces, '_existingNamespaces');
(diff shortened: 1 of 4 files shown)


lib/src/migrators/division.dart
```diff
@@ -187,6 +193,17 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
     }, isDivisionAllowed: true);
   }
 
+  /// Sets [_negatedParenthesized] to true when about to visit a negated
+  /// parenthesized expression.
+  @override
+  void visitUnaryOperationExpression(UnaryOperationExpression node) {
+    if (node.operator == UnaryOperator.minus &&
+        node.operand is ParenthesizedExpression) {
+      _negatedParenthesized = true;
```
nex3, line 202: You're effectively using this as an out-of-band way to pass an argument to `visitParenthesizedExpression()`... but what if instead you added a `{bool negated = false}` arg to `visitParenthesizedExpression()` and invoked it explicitly here?

jathak: Good point. Done.

nex3 (approved):

PR #196, opened 2021-06-04
Allow non-glob file path arguments containing glob syntax

faazshift:
Available examples and documentation don't specify that file path arguments should be passed in as quoted glob patterns to avoid processing by shells, rather than being able to be expanded by the shell before execution. In many cases this is fine, as patterns not containing glob pattern syntax will be handled as expected. However, if I pass a file like `assets/scss/pages/sport/\[sport\]/roster.module.scss` without quoting it, or use a matching glob pattern like `assets/**/*.scss` (also not quoted), expanded by a shell with globstar enabled, an error is thrown about `assets/scss/pages/sport/s` not existing.

With my change, if a passed argument is an existing file, it will be used without trying to handle it as a glob pattern.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.4.5
+
+* Glob syntax will no longer be resolved if a file with that literal name
+  exists.
+
 ## 1.4.4
 
 ### Division Migrator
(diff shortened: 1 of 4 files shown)


jathak (approved):
Thanks for this!

PR #197, opened 2021-06-07
Only ignore comments when finding insertion point

jathak:


jathak:
@Awjin, can you take a look?

Awjin (approved):

PR #198, opened 2021-06-18
Migrate to multiplication when possible

jathak:


lib/src/migrators/division.dart
```diff
@@ -295,6 +303,27 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
...
 
+  /// Given a division operation [node], patches it to use multiplication
+  /// instead if the reciprocal of the divisor can be accurately represented as
+  /// a decimal.
+  ///
+  /// Returns true if patched and false otherwise.
+  bool _tryMultiplication(BinaryOperationExpression node) {
+    if (!useMultiplication) return false;
+    if (node.right is! NumberExpression) return false;
+    var divisor = node.right as NumberExpression;
```
nex3, line 314:
```suggestion
    var divisor = node.right;
    if (divisor is! NumberExpression) return false;
```

This will let you avoid the `as`.

test/migrators/division/multiplication.hrx
```diff
@@ -0,0 +1,25 @@
...
+$b: $a / 4;
+$c: $a / 5;
+$d: ($a / 8);
+$e: ($a / 10);
+$f: ($a / 20);
+$g: $a/40;
+$h: $a/50;
+$i: $a/80;
+$j: ($a/100);
+$k: ($a/1000);
```
nex3, line 12: Test a division operation where the dividend is itself a mathematical operation—you'll need to be sure to keep it parenthesized in that case.

nex3 (approved):

dmix:
@jathak what was the rationale for this one? Just curious. 

It's it some CPU optimization? Preferring to not import the math lib?

jathak:
It's to avoid needing to load `sass:math` when not otherwise necessary and because we figured something like `$x * 0.5` was generally more readable than `math.div($x, 2)`.

PR #210, opened 2021-08-24
Use the new AST interfaces

jathak:
This updates the migrator to be compatible with the recent changes to
`DynamicImport.url` and `FunctionExpression.name`, and also uses the new
`SassReference` and `SassDeclaration` interfaces.

For now, it's still importing the Sass AST from `dart-sass`'s `src`,
rather than from the new `sass_api`.

### DIFF
--- a/lib/src/migrators/module/references.dart
+++ b/lib/src/migrators/module/references.dart
@@ -295,9 +295,7 @@ class _ReferenceVisitor extends RecursiveAstVisitor {
     var functions = _unresolvedReferences.keys.whereType<FunctionExpression>();
     for (var function in functions) {
       if (_isCssCompatibilityOverload(function)) continue;
-      var name = function.name;
-      if (name == null) continue;
-      var module = builtInFunctionModules[name];
+      var module = builtInFunctionModules[function.name];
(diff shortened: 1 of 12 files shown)


lib/src/migrators/module/references.dart
```diff
@@ -46,16 +45,14 @@ class References {
   ///
   /// Each value in this map must be a [VariableDeclaration] or an [Argument].
   final BidirectionalMap<MemberDeclaration<VariableDeclaration>,
-          MemberDeclaration<SassNode /*VariableDeclaration|Argument*/ >>
```
nex3, line 49: These shadow union types are still as informative now as they were before, right?

nex3 (approved):

lib/src/migrators/module/references.dart
jathak: Now that the actual getters used required are covered by `SassDeclaration`, I don't think there's any case where we actually have to do an instance check, so I thought it might be more readable to remove the union type comment and just specify what it is in the doc comment.

Would you prefer to keep the comments?

nex3: Either way is fine with me.

PR #217, opened 2021-12-02
Stop migrating division inside calc expressions

### LINKED ISSUE #216: Don't migrate division inside calc expressions
When using `sass-migrator division` against a file with the following expression:

```scss
margin-left: calc(var(--space-s-l) / -2);
margin-right: calc(var(--space-s-l) / -2);
```

It migrates it to:
```scss
margin-left: calc(math.div(var(--space-s-l), -2));
margin-right: calc(math.div(var(--space-s-l), -2));
```

Which then generates the following warning:
```
Warning: math.div() will only support number arguments in a future release.
Use list.slash() instead for a slash separator.

 margin-left: calc(math.div(var(--space-s-l), -2));
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

The same file had the following expression:

```scss
padding-left: calc(var(--space-s-l) / 2);
padding-right: calc(var(--space-s-l) / 2);
```

which was correctly changed to:

```scss
padding-left: calc(var(--space-s-l) * 0.5);
padding-right: calc(var(--space-s-l) * 0.5);
```

which makes it seem pretty clear that it was the negative number causing issues.

jathak:
Fixes #216.

### DIFF
--- a/lib/src/migrators/division.dart
+++ b/lib/src/migrators/division.dart
@@ -69,6 +69,10 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
   /// True when the current node is expected to evaluate to a number.
   var _expectsNumericResult = false;
 
+  /// True when the current node is in calc context, so division should be
+  /// left unchanged.
+  var _inCalcContext = false;
+
   /// The namespaces that already exist in the current stylesheet.
   Map<Uri, String?> get _existingNamespaces =>
       assertInStylesheet(__existingNamespaces, '_existingNamespaces');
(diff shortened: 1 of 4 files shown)


lib/src/migrators/division.dart
```diff
@@ -271,6 +292,11 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
   /// Returns true the `/` was migrated to either function call (indicating that
   /// parentheses surrounding this operation should be removed).
   bool _visitSlashOperation(BinaryOperationExpression node) {
+    if (_inCalcContext) {
+      node.left.accept(this);
+      node.right.accept(this);
+      return false;
+    }
```
nex3, line 299: Nit: newline here.

nex3 (approved):

PR #221, opened 2022-03-08
Remove a now unnecessary null assertion

natebosch:
In the latest version of `package:collection` the argument type for the
callback in `maxBy` was tightened since it has unnecessarily been marked
as nullable.

Remove the null assertion and bump the minimum version of `collection`.


jathak:
Thanks!

PR #228, opened 2022-09-07
Add media query migrator

jathak:
This PR adds a new `handleDeprecation` method to `Migrator`, which the new migrator uses instead of attempting to duplicate the logic from the parser.


lib/src/migrators/media_logic.dart
```diff
@@ -0,0 +1,56 @@
...
+    var visitor = _MediaLogicVisitor(
+        importCache, migrateDependencies, _expressionsToMigrate);
+    var result = visitor.run(stylesheet, importer);
+    missingDependencies.addAll(visitor.missingDependencies);
+    return result;
+  }
+}
+
+class _MediaLogicVisitor extends MigrationVisitor {
+  final Map<Uri, Set<FileSpan>> _expressionsToMigrate;
```
nex3, line 42: Nit: document this

nex3 (approved):

PR #245, opened 2023-11-21
Add calculation interpolation migrator

pamelalozano16:
Removes interpolation in calculation functions `calc()`, `clamp()`, `min()`, and `max()`.
More details on this lint [here](https://github.com/stylelint-scss/stylelint-scss/issues/539).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,5 +54,3 @@ review. In addition, because these models were trained indiscriminately and
 non-consensually on open-source code with a variety of licenses, it's not
 obvious that we have the moral or legal right to redistribute code they
 generate.
-
-
--- a/lib/src/migrators/calc_interpolation.dart
+++ b/lib/src/migrators/calc_interpolation.dart
@@ -1,0 +1,53 @@
+// Copyright 2023 Google LLC
+//
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import 'package:sass_api/sass_api.dart';
+
+import '../migration_visitor.dart';
+import '../migrator.dart';
+import '../patch.dart';
+
+/// Removes interpolation in calculation functions.
+class CalculationInterpolationMigrator extends Migrator {
+  final name = "calc-interpolation";
+  final description = r"Removes interpolation in calculation functions"
+      r"`calc()`, `clamp()`, `min()`, and `max()`";
+
+  @override
+  Map<Uri, String> migrateFile(
+      ImportCache importCache, Stylesheet stylesheet, Importer importer) {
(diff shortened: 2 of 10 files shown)


lib/src/runner.dart
```diff
@@ -18,6 +18,7 @@ import 'migrators/media_logic.dart';
 import 'migrators/module.dart';
 import 'migrators/namespace.dart';
 import 'migrators/strict_unary.dart';
+import 'migrators/calc_interpolation.dart';
```
jathak, line 21: Sort imports within a given section

See https://dart.dev/effective-dart/style#do-sort-sections-alphabetically

lib/src/runner.dart
```diff
@@ -59,6 +60,7 @@ class MigratorRunner extends CommandRunner<Map<Uri, String>> {
     addCommand(ModuleMigrator());
     addCommand(NamespaceMigrator());
     addCommand(StrictUnaryMigrator());
+    addCommand(CalculationInterpolationMigrator());
```
jathak, line 63: Sort these `addCommand` calls (so this new one should be first)

lib/src/migrators/calc_interpolation.dart
```diff
@@ -0,0 +1,51 @@
+// Copyright 2022 Google LLC
```
jathak, line 1: Nit: This should be 2023, since this file was first created in 2023.

lib/src/migrators/calc_interpolation.dart
```diff
@@ -0,0 +1,51 @@
...
+  Map<Uri, String> migrateFile(
+      ImportCache importCache, Stylesheet stylesheet, Importer importer) {
+    var visitor = _UnaryMigrationVisitor(importCache, migrateDependencies);
+    var result = visitor.run(stylesheet, importer);
+    missingDependencies.addAll(visitor.missingDependencies);
+    return result;
+  }
+}
+
+class _UnaryMigrationVisitor extends MigrationVisitor {
```
jathak, line 29: This name should be changed to something specific to this migrator (e.g. `_CalculationInterpolationVisitor`)

CONTRIBUTING.md
```diff
@@ -55,4 +55,7 @@ non-consensually on open-source code with a variety of licenses, it's not
 obvious that we have the moral or legal right to redistribute code they
 generate.
 
-
+## Testing
```
jathak, line 58: Item 1 here is already stated [here](https://github.com/sass/migrator/blob/main/test/migrators/README.md). It probably makes sense to add the instructions for running tests to that file too.

pamelalozano16: Done! Thanks for the review

jathak (approved):

PR #248, opened 2023-12-09
Add parenthesis in operations to keep order.

pamelalozano16:
Bug fix: Removing interpolation was causing changes in the order of calculations.
Replacing interpolation with parenthesis in these cases produces the same result.

Ex. `$a * #{$b + $c}` is not the same calculation as `$a * $b + $c`, but it is the same as .`$a * ($b + $c)`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.8.2
+
+### Calc Functions Interpolation Migrator
+
+* Add parentheses in place of interpolation when necessary to preserve the evaluation order.
+
 ## 1.8.1
 
 ### Calc Functions Interpolation Migrator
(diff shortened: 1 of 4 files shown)


test/migrators/calc_interpolation/calc_remove_interpolation.hrx
```diff
@@ -2,10 +2,14 @@
 $b: 10;
 $c: 1;
 $d: 5;
-.a { .b: calc($b - #{$c + 1}); }
+//Single interpolation
```
jathak, line 5: Add a space between the `//` and the text here, and a blank line between this and the previous section.

jathak (commented):
Please add a changelog entry and bump the pubspec version to 1.8.2-dev (since we can release this and #249 at the same time)

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 1.8.2
+
+### Calc Functions Interpolation Migrator
+
+* Add parenthesis in operations to keep order bug fix.
```
jathak, line 5: A clearer wording here would probably be something like "Add parentheses in place of interpolation when necessary to preserve the evaluation order"

jathak (approved):

PR #249, opened 2023-12-12
Detect dependency loops in module migrator

### LINKED ISSUE #137: Migrator can generate @use loops
If the entrypoint imports a file that uses variables that were originally defined in the entrypoint, the migrator generates a `@use` that imports the original file. For example:

```scss
// entrypoint.scss
$var: value;

@import "other";
```

```scss
// _other.scss
@debug $value;
```

is migrated to:

```scss
// entrypoint.scss
$var: value;

@use "other";
```

```scss
// _other.scss
@use "entrypoint";
@debug entrypoint.$value;
```

This is invalid, as it creates a dependency loop. Since it's not totally clear what the user intended, the migrator should probably error out.

pamelalozano16:
Fix #137.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.5.8
+
+* Detect dependency loops in module migrator fix.
+
 ## 2.5.7
 
 * Fix a bug where the `--migrate-deps` flag would not apply to dependencies
--- a/lib/src/migrators/module.dart
+++ b/lib/src/migrators/module.dart
@@ -16,6 +16,7 @@ import '../migration_visitor.dart';
 import '../migrator.dart';
 import '../patch.dart';
 import '../utils.dart';
+import '../util/dependency_graph.dart';
 import '../util/member_declaration.dart';
 import '../util/node_modules_importer.dart';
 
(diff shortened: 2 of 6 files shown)


lib/src/migrators/module.dart
```diff
@@ -200,6 +200,21 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+  void _addDependency(Uri source, Uri importedPath) {
+    if (_dependencies.containsKey(importedPath) &&
+        _dependencies[importedPath] == source) {
+      // Throw an error indicating a potential loop.
+      var sourceUrl = _absoluteUrlToDependency(source);
+      var importedPathUrl = _absoluteUrlToDependency(importedPath);
+      throw MigrationException(
+          'Dependency loop detected: ${sourceUrl.item1} -> ${importedPathUrl.item1}');
+    }
+    _dependencies[source] = importedPath;
```
jathak, line 215: Given this only assigns one imported path per source (overriding any previous ones), I think this could have issues if one file imports multiple others

test/migrators/module/namespace_references/loop_error.hrx
```diff
@@ -0,0 +1,21 @@
+<==> arguments
```
jathak, line 1: While it makes sense to error in a case like this, it's not really the core of the issue in #137, since the code does not compile as-is, since `$var` is referenced before it's defined.

We specifically care about cases where the code compiles before migration due to an implicit dependency, but creates a loop when we try to migrate it to `@use` (see the example in Natalie's original comment on #137).

jathak (changes requested):

lib/src/migrators/module.dart
pamelalozano16: I changed the data structure to a dependency graph, this way it's easier to escalate and we can detect any loop in the imports.

test/migrators/module/namespace_references/loop_error.hrx
pamelalozano16: I reworded the error to give fixing options rather than just pointing out the error, also added Natalie's example in the tests.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 2.0.4
+* Detect dependency loops in module migrator fix.
```
jathak, line 3: I think a clearer way to describe this change to end users is something like:

> Previously, migrating certain files with implicit dependencies on one another could create dependency loops. This fixes the migrator to instead error when it encounters these situations, as a working migration is not possible without some sort of manual refactoring.

Also, put this under a "Module Migrator" header

lib/src/migrators/module.dart
```diff
@@ -199,6 +200,33 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
   /// The values of the --forward flag.
   final Set<ForwardType> forwards;
 
+  /// Dependencies where keys represent source URIs and values represent imported URIs.
+  final DependencyGraph _dependencies = DependencyGraph();
```
jathak, line 244: Nit: The type here is unnecessary since it can be inferred from the initializer.

```suggestion
  final _dependencies = DependencyGraph();
```

test/migrators/module/namespace_references/loop_error_one_reference.hrx
```diff
@@ -0,0 +1,24 @@
...
+<==> input/_library.scss
+a {
+  margin: $var1;
+}
+
+<==> error.txt
+Error: Dependency loop detected: entrypoint -> library.
+To resolve this issue, consider either of the following:
+1. Remove the import statement that causes the dependency loop.
+2. Declare the variables used in the other stylesheet within the same stylesheet.
```
jathak, line 17: How does this handle functions and mixins?

PR #250, opened 2023-12-18
Keep interpolation in var CSS function

pamelalozano16:
Bug fix: Keep interpolation within `var()` functions.

### DIFF
--- a/lib/src/migrators/calc_interpolation.dart
+++ b/lib/src/migrators/calc_interpolation.dart
@@ -37,19 +37,28 @@ class _CalculationInterpolationVisitor extends MigrationVisitor {
     const calcFunctions = ['calc', 'clamp', 'min', 'max'];
     final interpolation = RegExp(r'\#{\s*[^}]+\s*}');
     final hasOperation = RegExp(r'[-+*/]+');
+    final isVarFunc = RegExp(
+        r'var\(#{[a-zA-Z0-9#{$}-]+}\)|var\(\-\-[a-zA-Z0-9\$\#\{\}\-]+\)');
     if (calcFunctions.contains(node.name)) {
       for (var arg in node.arguments.positional) {
         var newArg = arg.toString();
-        for (var match in interpolation.allMatches(arg.toString())) {
-          var noInterpolation =
-              match[0].toString().substring(2, match[0].toString().length - 1);
+        var varFuncArgs = isVarFunc.allMatches(newArg);
+        if (varFuncArgs.isNotEmpty) {
+          newArg = newArg.replaceAll(isVarFunc, 'var()');
+        }
+
+        for (var match in interpolation.allMatches(newArg)) {
+          var noInterpolation = match[0]!.substring(2, match[0]!.length - 1);
           if (hasOperation.hasMatch(noInterpolation)) {
             noInterpolation = '(' + noInterpolation + ')';
           }
-          newArg = newArg
-              .toString()
-              .replaceAll(match[0].toString(), noInterpolation);
+          newArg = newArg.toString().replaceAll(match[0]!, noInterpolation);
         }
+
+        for (var match in varFuncArgs) {
+          newArg = newArg.replaceFirst('var()', match[0]!);
+        }
+
         if (newArg != arg.toString()) {
           var interpolationSpan =
               node.span.file.span(arg.span.start.offset, arg.span.end.offset);
--- a/test/migrators/calc_interpolation/calc_remove_interpolation.hrx
+++ b/test/migrators/calc_interpolation/calc_remove_interpolation.hrx
@@ -21,6 +21,9 @@ $d: 5;
 // CSS Custom properties keep interpolation
 .a { --test: calc(#{$b} + 1);}
 
+// var() nested
+.a { .b: calc(var(#{$b}) + #{$c + 2} + var(--a-#{$d}-b)); }
+
 <==> output/entrypoint.scss
 $b: 10;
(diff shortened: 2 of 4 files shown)


lib/src/migrators/calc_interpolation.dart
```diff
@@ -37,10 +37,16 @@ class _CalculationInterpolationVisitor extends MigrationVisitor {
     const calcFunctions = ['calc', 'clamp', 'min', 'max'];
     final interpolation = RegExp(r'\#{\s*[^}]+\s*}');
     final hasOperation = RegExp(r'[-+*/]+');
+    final isVarFunc = RegExp(r'var\(\#{\s*[^}]+\s*}\)');
```
jathak, line 40: 1. I don't think the `#` needs to be escaped here.
2. This regex won't match something like `var(--prefix-#{$variable}-suffix)`

lib/src/migrators/calc_interpolation.dart
```diff
@@ -37,10 +37,16 @@ class _CalculationInterpolationVisitor extends MigrationVisitor {
...
     if (calcFunctions.contains(node.name)) {
       for (var arg in node.arguments) {
         var newArg = arg.toString();
-        for (var match in interpolation.allMatches(arg.toString())) {
+        var varFuncArgs;
+        if (isVarFunc.hasMatch(newArg)) {
+          varFuncArgs = isVarFunc.allMatches(newArg);
+          newArg = newArg.replaceAll(isVarFunc, 'var()');
+        }
+        for (var match in interpolation.allMatches(newArg)) {
```
jathak, line 49: If you're declaring an uninitialized variable, you need to give it an explicit type (in this case, it's `Iterable<RegExpMatch>?`).

Alternatively, you could just always set `varFuncArgs` to the result of calling `allMatches` and then check `varFuncArgs.isNotEmpty` instead of calling `hasMatch`.

test/migrators/calc_interpolation/calc_remove_interpolation.hrx
```diff
@@ -18,6 +18,9 @@ $d: 5;
 // Nested and more interpolations
 .a { .b: calc(#{$b} + max(#{$c, 2})); }
 
+// var() nested
+.a { .b: calc(var(#{$b}) + #{$c + 2} + var(#{$d})); }
+
```
jathak, line 23: Add a test for something like  `var(--prefix-#{$variable}-suffix)` mentioned above.

jathak (commented):
You'll also want to merge/rebase this on top of main, as I needed to make some changes to make this compatible with the latest `sass_api`

lib/src/migrators/calc_interpolation.dart
pamelalozano16: Thanks! Done

test/migrators/calc_interpolation/calc_remove_interpolation.hrx
pamelalozano16: 👍

lib/src/migrators/calc_interpolation.dart
pamelalozano16: 👍

lib/src/migrators/calc_interpolation.dart
```diff
@@ -50,6 +55,11 @@ class _CalculationInterpolationVisitor extends MigrationVisitor {
               .toString()
               .replaceAll(match[0].toString(), noInterpolation);
         }
+
+        for (var match in varFuncArgs) {
+          newArg = newArg.replaceFirst('var()', match[0].toString());
```
jathak, line 60: `match[0]` is typed as `String?`, but it's never actually null, so I think it makes more sense to write this as `match[0]!` (both here and in some of the existing code above)

test/migrators/calc_interpolation/calc_remove_interpolation.hrx
```diff
@@ -37,3 +40,6 @@ $d: 5;
 
 // Nested and more interpolations
 .a { .b: calc($b + max($c, 2)); }
+
+// var() nested
+.a { .b: calc(var(#{$b}) + #{$c + 2} + var(--a-#{$d}-b)); }
```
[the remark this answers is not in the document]
pamelalozano16: Yes, there was a bug in the Regex! Thanks for catching that

lib/src/migrators/calc_interpolation.dart
```diff
@@ -37,19 +37,30 @@ class _CalculationInterpolationVisitor extends MigrationVisitor {
...
             noInterpolation = '(' + noInterpolation + ')';
           }
-          newArg = newArg
-              .toString()
-              .replaceAll(match[0].toString(), noInterpolation);
+          newArg = newArg.toString().replaceAll(matchRegex, noInterpolation);
         }
+
+        for (var match in varFuncArgs) {
+          String matchRegex = match[0].toString();
```
jathak, line 60: Factoring this out into a local variable isn't necessary (especially since it's only used once here). My previous comment was saying to use the [non-null assertion operator (`!`)](https://dart.dev/language/operators#other-operators) instead of `toString`, since `match[0]` will never be null, it's just typed as `String?`.

pamelalozano16: Ohh gotcha, I misread the comment sorry about that. Thanks ! :)

jathak (approved):
The version in the pubspec also needs to be bumped to 2.0.1, otherwise this LGTM

PR #251, opened 2024-01-03
Use latest Dart Sass and release migrator 2.0.0

jathak:
- Deleted the obsolete `media-logic` migrator (breaking change bumps version to 2.0.0).
- Moved off of the tuple package to native Dart tuples.
- Fixed the `calc-interpolation` and `division` migrators to eliminate their use of the removed `CalculationExpression` AST node.
- Created a new `ScopedAstVisitor` that uses `Scope` to track Sass member declarations. `MigrationVisitor` and `_ReferenceVisitor` now extend this.
- Refactored the `division` migrator to use patterns.


lib/src/migrators/division.dart
```diff
@@ -244,40 +272,45 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
...
-    }
-    if (channels == null ||
-        channels.hasBrackets ||
-        channels.separator != ListSeparator.space ||
-        channels.contents.length != 3 ||
-        channels.contents.last is! BinaryOperationExpression) {
-      return false;
-    }
+    var channels = switch (node.arguments) {
+      ArgumentInvocation(positional: [ListExpression arg, ...], named: Map()) =>
```
nex3, line 276: Should this be `[ListExpression arg]`? The old code only matches if there's only one positional argument.

Also, I think `named: Map()` just checks that `named`'s type is `Map`. If you want to check that it's empty, you'll need to do `named: Map(length: 0)`.

lib/src/migrators/division.dart
```diff
@@ -244,40 +272,45 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
...
+          hasBrackets: false,
+          separator: ListSeparator.space,
+          contents: [_, _, BinaryOperationExpression last]
+        )) {
+      // Handles cases like `rgb(10 20 30/2 / 0.5)`, since converting `30/2`
+      // to `div(30, 20)` would cause `/ 0.5` to be interpreted as division.
+      switch (last) {
+        case BinaryOperationExpression(
+            left: NumberExpression _,
+            right: NumberExpression _
```
nex3, line 297: Here I think `NumberExpression()` is more idiomatic

lib/src/migrators/division.dart
```diff
@@ -244,40 +272,45 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
...
+          ):
+          break;
+        default:
+          _patchSpacesToCommas(channels);
+          _patchOperatorToComma(last);
+      }
```
nex3, line 303: This could just be:

```suggestion
      if (last case BinaryOperationExpression(
        left: NumberExpression(),
        right: NumberExpression()
      )) {
        break;
      }
      _patchSpacesToCommas(channels);
      _patchOperatorToComma(last);
```

lib/src/migrators/division.dart
```diff
@@ -330,110 +366,81 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
...
-    var operatorSpan = node.left.span
-        .extendThroughWhitespace()
-        .end
-        .pointSpan()
-        .extendIfMatches('/');
-    addPatch(Patch(operatorSpan, '*'));
-    addPatch(Patch(node.right.span, '${1 / divisor.value}'));
-    return true;
+    if (node.right case NumberExpression(unit: null, value: var divisor)) {
+      if (!_allowedDivisors.contains(divisor)) return false;
```
nex3, line 370: I'd put this in a `when` clause since it's logically part of the same condition as the `if`.

lib/src/migrators/division.dart
```diff
@@ -330,110 +366,81 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
...
 
   /// Returns true if [node] is entirely composed of number literals and slash
   /// operations.
   bool _onlySlash(Expression node) {
-    if (node is NumberExpression) return true;
-    if (node is BinaryOperationExpression) {
-      return node.operator == BinaryOperator.dividedBy &&
-          _onlySlash(node.left) &&
-          _onlySlash(node.right);
+    switch (node) {
```
nex3, line 417: I'd make this a switch expression

lib/src/migrators/division.dart
```diff
@@ -330,110 +366,81 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
...
 
   /// Returns true if [node] contains an interpolation.
   bool _containsInterpolation(Expression node) {
-    if (node is ParenthesizedExpression) {
-      return _containsInterpolation(node.expression);
-    }
-    if (node is BinaryOperationExpression) {
-      return _containsInterpolation(node.left) ||
-          _containsInterpolation(node.right);
+    switch (node) {
```
nex3, line 433: This too

lib/src/migrators/division.dart
```diff
@@ -330,110 +366,81 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
...
+      case ParenthesizedExpression(:var expression):
+      case UnaryOperationExpression(operand: var expression):
```
nex3, line 435: I generally try to use `||` rather than cascading `case`s because it's easier to translate to and from expression form

lib/src/migrators/division.dart
```diff
@@ -456,9 +463,8 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
   /// ParenthesizedExpression.
   void _patchParensIfAny(SassNode node) {
     if (node is! ParenthesizedExpression) return;
-    var expression = node.expression;
-    if (expression is BinaryOperationExpression &&
-        expression.operator == BinaryOperator.dividedBy) {
+    if (node.expression
+        case BinaryOperationExpression(operator: BinaryOperator.dividedBy)) {
```
nex3, line 467: I'd fold these together into a single `if`/`case`

lib/src/migrators/division.dart
```diff
@@ -482,3 +488,41 @@ class _DivisionMigrationVisitor extends MigrationVisitor {
...
+/// number, or unknown.
+enum _NumberStatus {
+  yes,
+  no,
+  maybe;
+
+  /// Returns [yes] if [node] is definitely a number, [no] if [node] is
+  /// definitely not a number, and [maybe] otherwise.
+  static _NumberStatus of(Expression node) {
+    switch (node) {
```
nex3, line 502: Another good candidate for expression form

lib/src/migrators/module.dart
```diff
@@ -310,24 +309,24 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
     // that the import-only file still includes its CSS.
-    var dependency =
-        _absoluteUrlToDependency(entrypoint, relativeTo: importOnlyUrl).item1;
+    var (dependency, _) =
+        _absoluteUrlToDependency(entrypoint, relativeTo: importOnlyUrl);
     var forwards = forwardsByUrl.remove(entrypoint);
     var entrypointForwards = forwards != null
         ? _forwardRulesForShown(entrypoint, '"$dependency"', forwards, {})
         : ['@forward "$dependency"'];
     var tuples = [
```
nex3, line 318: Why have a separate variable for this rather than looping over `forwardsByUrl` directly in the `forwardLines` initializer?

lib/src/migrators/module.dart
```diff
@@ -379,9 +378,9 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
         .toSet()) {
       if (url == currentUrl || _forwardedUrls.contains(url)) continue;
       var forwards =
-          _makeForwardRules(url, '"${_absoluteUrlToDependency(url).item1}"');
+          _makeForwardRules(url, '"${_absoluteUrlToDependency(url).$1}"');
       if (forwards == null) continue;
-      var isRelative = _absoluteUrlToDependency(url).item2;
+      var (_, isRelative) = _absoluteUrlToDependency(url);
```
nex3, line 383: Since you're accessing both halves of `_absoluteUrlToDependency()` here, you might as well call it early and assign them both to variables

lib/src/migrators/module.dart
```diff
@@ -771,17 +768,17 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
       Uri? ruleUrl = import.url;
       var tuple = importCache.canonicalize(ruleUrl,
           baseImporter: importer, baseUrl: currentUrl, forImport: true);
```
nex3, line 770: Destructure this tuple

lib/src/migrators/module.dart
```diff
@@ -1272,11 +1264,11 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
   /// The first item of the returned tuple is the dependency, the second item
   /// is true when this dependency is resolved relative to the current URL and
   /// false when it's resolved relative to a load path.
-  Tuple2<Uri, bool> _absoluteUrlToDependency(Uri url, {Uri? relativeTo}) {
+  (Uri, bool) _absoluteUrlToDependency(Uri url, {Uri? relativeTo}) {
     relativeTo ??= currentUrl;
     var tuple = _originalImports[url];
-    if (tuple != null && tuple.item2 is NodeModulesImporter) {
-      return Tuple2(tuple.item1, false);
+    if (tuple case (var url, NodeModulesImporter _)) {
```
nex3, line 1270: Nit: I'd inline the variable here.

lib/src/util/scoped_ast_visitor.dart
```diff
@@ -0,0 +1,101 @@
...
+import 'package:sass_api/sass_api.dart';
+
+import 'member_declaration.dart';
+import 'scope.dart';
+
+/// A recursive AST visitor that tracks the Sass members declared in the
+/// current scope.
+abstract class ScopedAstVisitor
+    with RecursiveStatementVisitor, RecursiveAstVisitor {
+  /// The current scope, containing any visible without a namespace to the
```
nex3, line 17: Any visible what?

lib/src/util/scoped_ast_visitor.dart
```diff
@@ -0,0 +1,101 @@
...
+
+  /// Creates a new child scope, declares [node]'s arguments within it and
+  /// then visits [node]'s children.
+  @override
+  void visitCallableDeclaration(CallableDeclaration node) {
+    scoped(() {
+      for (var argument in node.arguments.arguments) {
+        currentScope.variables[argument.name] = MemberDeclaration(argument);
+        var defaultValue = argument.defaultValue;
+        if (defaultValue != null) visitExpression(defaultValue);
```
nex3, line 68: Nit: Prefer `if`/`case` for situations where a variable is only used within a null-check block, so the variable's scope matches its usage

lib/src/migration_visitor.dart
```diff
@@ -117,26 +118,31 @@ abstract class MigrationVisitor
...
       //
       // TODO(nweiz): Add a flag to override this behavior for load paths
       // (#104).
-      if (result.item1 != _importer) return;
+      if (newImporter != _importer) return;
 
       var oldImporter = _importer;
-      _importer = result.item1;
-      var stylesheet = result.item2;
+      var oldScope = currentScope;
```
nex3, line 140: Consider adding a `when:` parameter to `scoped()` and using that here.

nex3 (changes requested):

lib/src/migrators/module.dart
jathak: Mostly just an artifact of incrementally moving from however this was originally written -> collection literals -> patterns.

I've updated this to combine them. Is using a single-option switch expression to extract `ruleUrl` a good idea here, or should I just inline the `_absoluteUrlToDependency` call?

lib/src/migrators/module.dart

lib/src/migrators/division.dart
jathak: Fixed

lib/src/migrators/division.dart
jathak: Fixed

lib/src/migrators/division.dart
jathak: The break was to break out of the switch, since we only want to call the `_patch*` functions when the pattern is _not_ matched, but we do still want to do the `_withContext` part either way. Is this structure (with an empty `if` and the patch calls in an `else`) better than the switch?

lib/src/migrators/division.dart
jathak: Done. Forgot that when clauses where an option.

lib/src/migrators/division.dart
jathak: Done

lib/src/migrators/division.dart
jathak: Done

lib/src/migrators/division.dart
jathak: Done

lib/src/migrators/division.dart
jathak: I don't think a single if/case is possible here, since we want to return in both the case where `node` is a parenthesized division expression and the case where it's not parenthesized. Does it make sense to leave this as-is or convert to a switch statement?

lib/src/migrators/division.dart
jathak: Done

lib/src/migrators/division.dart
nex3: Oh, I see. In Dart 3 you no longer need a `break` at the end of a `case` block; they don't fall through unless either the block is empty or you explicitly opt into it with `continue`.

In this case I'd like `if`/`case`/`else`.

lib/src/migrators/division.dart
nex3: Oh, I see what you mean. Yeah in this case I'd use a `switch` I think, since you're effectively enumerating cases.

lib/src/migrators/module.dart
nex3: I'd just inline the call. I usually only resort to fake-`let` patterns if I need to re-use the variable (also, I'd use `if`/`case` over `switch` for that).

nex3 (approved):

lib/src/migrators/division.dart
jathak: What is the idiomatic way to have an empty `case` block that doesn't fall through?

nex3: In _that_ case you need `break`, but only that case.

PR #254, opened 2024-01-16
Fix operators regex to remove unnecessary parenthesis in SCSS variables.

pamelalozano16:
`calc(#{$a-b})` was being migrated to `calc(($a-b))` because of the `-` less operator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2.0.2
+
+### Calc Functions Interpolation Migrator
+
+* Fix the interpretation of a dash in a variable name as a minus sign.
+
 ## 2.0.1
 
 ### Calc Functions Interpolation Migrator
--- a/test/migrators/calc_interpolation/calc_remove_interpolation.hrx
+++ b/test/migrators/calc_interpolation/calc_remove_interpolation.hrx
@@ -1,4 +1,5 @@
 <==> input/entrypoint.scss
+$a-b: 5;
 $b: 10;
 $c: 1;
 $d: 5;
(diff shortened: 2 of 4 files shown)


test/migrators/calc_interpolation/calc_remove_interpolation.hrx
```diff
@@ -9,14 +9,14 @@ $d: 5;
...
-    .c: calc(100% - #{$table_title_height + 2px});
+    .c: calc(100% - #{$TABLE_TITLE + 2px});
 }
 
 // Nested
 .a { .b: calc(3 + max(#{$c, 2})); }
 
 // Nested and more interpolations
-.a { .b: calc(#{$b} + max(#{$c, 2})); }
+.a { .b: calc(#{$a-b} + max(#{$c, 2})); }
```
jathak, line 19: Add a declaration for `$a-b`. The migrator _can_ run on files that parse but don't compile, but it's best if our test cases are all valid Sass code the parses and compiles.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.0.2
```
jathak, line 1: The version in the pubspec should always match the latest version in the changelog (or have the `-dev` suffix if you're not planning to release that fix immediately)

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
...
+
+### Calc Functions Interpolation Migrator
+
+* Fix operators regex to remove unnecessary parenthesis in SCSS variables.
```
jathak, line 5: This describes an implementation detail rather than the actual user-visible fix. This should instead say something about fixing it so it no longer interprets a dash in a variable name as a minus sign.

CHANGELOG.md
pamelalozano16: The version of the pubspec is matched to `2.0.2`.. am I missing something?

CHANGELOG.md
pamelalozano16: Thanks!

CHANGELOG.md
jathak: Oh, whoops. My bad. Somehow missed it

jathak (approved):

PR #259, opened 2024-07-17
Add --built-in-only flag to module migrator

jathak:
Also fixes some built-in functions that weren't being migrated.


lib/src/migrators/module.dart
```diff
@@ -776,7 +791,12 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
         }
       }
       if (ruleUrl != null) {
-        if (_useAllowed) {
+        if (builtInOnly) {
+          _upstreamStylesheets.add(currentUrl);
+          if (migrateDependencies) visitDependency(ruleUrl, import.span);
+          _upstreamStylesheets.remove(currentUrl);
+          return;
```
nex3, line 798: Why return eagerly here? What if there are multiple imports in a single `@import` that all need to be traversed?

test/migrators/module/built_in_functions/builtin_only.hrx
nex3, line 1: There are a lot of special cases for `builtInOnly` in the module migrator. It could probably use a few more tests to exercise more of them.

nex3 (changes requested):

test/migrators/module/built_in_functions/builtin_only.hrx
jathak: Done

lib/src/migrators/module.dart
jathak: Good catch. Fixed

nex3 (commented):
Consider naming the flag `--built-in-only`, since "built-in" is usually spelled with a hyphen.

lib/src/migrators/module.dart
```diff
@@ -613,7 +630,7 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
       });
     }
 
-    if (node.name == "get-function") {
+    if (node.name == "get-function" && !builtInOnly) {
```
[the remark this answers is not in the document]
jathak: Fixed

lib/src/migrators/module.dart
```diff
@@ -776,14 +793,19 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
         }
       }
       if (ruleUrl != null) {
-        if (_useAllowed) {
+        if (builtInOnly) {
+          _upstreamStylesheets.add(currentUrl);
+          if (migrateDependencies) visitDependency(ruleUrl, import.span);
+          _upstreamStylesheets.remove(currentUrl);
```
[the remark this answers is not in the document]

nex3 (approved):

PR #260, opened 2024-09-18
Add legacy color function migrator

jathak:


test/migrators/color/global.hrx
```diff
@@ -0,0 +1,39 @@
...
+  n: fade-in(gold, 0.1);
+  o: lighten(gold, 10%);
+  p: darken(gold, 10%);
+}
+
+<==> output/entrypoint.scss
+@use "sass:color";
+
+a {
+  b: color.channel(gold, 'red', rgb);
```
nex3, line 24: Can we have an explicit `$space: ` on these as well?

nex3 (changes requested):

test/migrators/color/global.hrx
jathak: Done

nex3 (approved):

PR #267, opened 2024-12-18
Support --pkg-importer=node

### LINKED ISSUE #264: Is it possible to handle package importers when using the sass-migrator?
If I try to use the sass-migrator on a codebase that has been convert to using `--pkg-importer=node`, then I get an error whenever I hit one of those statements, like such:

```
Error: Could not find Sass file at 'pkg:bourbon'.

5 │ @import 'pkg:bourbon'
```

It doesn't look like there is a `--pkg-importer` option for the sass-migrator. So far what I've been doing is moving these imports back to the `--load-path` version, running the migrator, then moving them back to the package migrator, but it's a bit tedious to do so.

Thanks for all the work on this tool by the way! It has simplified a lot of things for us!

jathak:
Fixes #264


lib/src/migrator.dart
```diff
@@ -68,9 +68,11 @@ abstract class Migrator extends Command<Map<Uri, String>> {
   Map<Uri, String> run() {
     var allMigrated = <Uri, String>{};
     var importer = FilesystemImporter('.');
-    var importCache = ImportCache(
-        importers: [NodeModulesImporter()],
-        loadPaths: globalResults!['load-path']);
+    var importCache = ImportCache(importers: [
+      NodeModulesImporter(),
+      if ((globalResults!['pkg-importer'] as List<String>).contains('node'))
```
nex3, line 73: I think it would communicate the likely extensions of this behavior better to make this a for loop.

nex3 (approved):

Issue #273, opened 2025-04-08
Some variables get `--` replaced to `-`


Tofandel:
When using the migrator with `npx sass-migrator module --migrate-deps src/theme/theme.scss`

Some variables get renamed

Example

```diff

- $-color-success-light: color.mix($-color-white, $-color-success, 80%) !default;
+ $--color-success-light: mix($--color-white, $--color-success, 80%) !default;
- $-color-warning-light: color.mix($-color-white, $-color-warning, 80%) !default;
+ $--color-warning-light: mix($--color-white, $--color-warning, 80%) !default;
- $-color-danger-light: color.mix($-color-white, $-color-danger, 80%) !default;
+ $--color-danger-light: mix($--color-white, $--color-danger, 80%) !default;
- $--color-info-light: mix($-color-white, $-color-info, 80%) !default;
- $--color-info-light: color.mix($-color-white, $-color-info, 80%) !default;
```

I do not want this renaming to happen, I tried with `--remove-prefix=app` but that changed nothing

What is causing this?

Tofandel:
Related issue running the migrator with

npx sass-migrator module --migrate-deps src/theme/theme.scss --remove-prefix=app
```
$app--color-black: #000000 !default;
```

Results in 
```
$a-color-black: #000000 !default;
```

Instead of the expected `$--color-black`

PR #274, opened 2025-04-08
fix: private variable detection to not match multiple dashes

Tofandel:
Partial fix for #273

This is not a fix for the conflicting patch that seems to happen though

```scss
$app-color-black: #000000;
```

npx sass-migrator module test.scss --remove-prefix=app

```scss
$acolor-black: #000000;
```

So I will draft until I find the cause of this


nex3:
Superseding this with https://github.com/sass/migrator/pull/275 so I can get a review from a colleague

PR #275, opened 2025-04-09
Fix the conversion of private names to public names

nex3:
This handles cases where the private names start with multiple dashes or underscores or are even composed entirely of dashes and underscores. It also fixes bugs where names starting with underscores weren't considered private.

Partially addresses https://github.com/sass/migrator/issues/273


Tofandel:
It's up for debate that dashes in front are private variables, it's just a convention and not a hard rule. As such this behavior needs an option to be able to be turned off

lib/src/migrators/module.dart
```diff
@@ -352,11 +356,11 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+    if (_isPrivate(name) &&
         references.referencedOutsideDeclaringStylesheet(declaration)) {
-      // Remove leading `-` since private members can't be accessed outside
-      // the module they're declared in.
-      name = name.substring(1);
+      // Private members can't be accessed outside the module they're declared
+      // in.
+      name = _privateToPublic(name);
     }
     name = _unprefix(name);
```
Tofandel, line 365: I believe the unprefixing should be done before the       `name = _privateToPublic(name);` to avoid the other issue I outlined

I am not sure what would cause the other issue but I believe a conflicting patch

lib/src/migrators/module.dart
```diff
@@ -1228,11 +1252,16 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
-    var unprivateName = isPrivate ? name.substring(1) : name;
+    var isPrivate = _isPrivate(name);
+    var unprivateName = isPrivate ? _privateToPublic(name) : name;
     var prefix = _prefixFor(unprivateName);
     if (prefix == null) return name;
-    return (isPrivate ? '-' : '') + unprivateName.substring(prefix.length);
+
+    var withoutPrefix = unprivateName.substring(prefix.length);
+    if (_isPrivate(withoutPrefix)) {
+      withoutPrefix = _privateToPublic(withoutPrefix);
```
Tofandel, line 1262: In this case maybe it's better to only run substring(1), otherwise multiple dashes will get replaced to 1 dash still

test/migrators/module/scope/pseudoprivate/double_dash.hrx
```diff
@@ -0,0 +1,19 @@
...
+$--var: red;
+
+<==> output/entrypoint.scss
+@use "library";
+a {
+  color: library.$var;
+}
+
+<==> output/_library.scss
+$var: red;
```
Tofandel, line 19:
```suggestion
$--red: red
$var: $--red;
```

test/migrators/module/scope/pseudoprivate/underscore.hrx
```diff
@@ -0,0 +1,19 @@
+<==> arguments
+--migrate-deps
+<==> input/entrypoint.scss
+@import "library";
+a {
+  color: $_var;
+}
+
+<==> input/_library.scss
+$_var: red;
```
Tofandel, line 10:
```suggestion
$_red: red;
$_var: $_red;
```

Same here

lib/src/migrators/module.dart
```diff
@@ -1228,11 +1252,16 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+    }
+    return (isPrivate ? '-' : '') + withoutPrefix;
```
Tofandel, line 1264:
```suggestion
    return (isPrivate ? name.substring(0, 1)) + withoutPrefix;
```

Tofandel:
I added suggested tests and fixed the reported behaviors of remove-prefix in #276

lib/src/migrators/module.dart
```diff
@@ -365,6 +369,27 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+    return identifier.startsWith('-') || identifier.startsWith('_');
+  }
+
+  /// Converts a private identifier to a public one.
+  String _privateToPublic(String identifier) {
+    assert(_isPrivate(identifier));
+    for (var i = 0; i < identifier.length; i++) {
+      var char = identifier.codeUnitAt(i);
+      if (char != $dash && char != $underscore) {
+        return identifier.substring(i);
```
pamelalozano16, line 385: What about `identifier.replace(Regex(r"^-|^_"),"")` ?

lib/src/migrators/module.dart
```diff
@@ -365,6 +369,27 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+  ///
+  /// Assumes [identifier] is a valid CSS identifier.
+  bool _isPrivate(String identifier) {
+    return identifier.startsWith('-') || identifier.startsWith('_');
+  }
+
+  /// Converts a private identifier to a public one.
+  String _privateToPublic(String identifier) {
+    assert(_isPrivate(identifier));
+    for (var i = 0; i < identifier.length; i++) {
```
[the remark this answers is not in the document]
Tofandel: It's just iterating each char until it finds a character that's not an underscore or dash, then returns the string without the iterated part

This is more performant than regex

lib/src/migrators/module.dart
Tofandel: The correct one would be `identifier.replace(Regex(r"^[-_]+"),"")`

Tofandel: And if the returned string is empty then return a generated var

nex3: I habitually avoid regular expressions because they're somewhat more expensive than explicit parsing. If you'd prefer it, I can make the change, but I think the current way does work.

lib/src/migrators/module.dart
nex3: Why? `_privateToPublic()` should handle multiple dashes gracefully.

lib/src/migrators/module.dart
nex3: Generally I'm okay with leaning towards encouraging `-`, since that's definitely the more CSS-style prefix.

lib/src/migrators/module.dart
```diff
@@ -1227,12 +1248,33 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+      prefix = _prefixFor(unprivateName);
+      if (prefix == null)
+        return isPrivate && forcePublic ? unprivateName : name;
+      withoutPrefix = unprivateName.substring(prefix.length);
+    }
+
+    if (_isPrivate(withoutPrefix)) {
+      withoutPrefix = _privateToPublic(withoutPrefix);
+    }
+    return (isPrivate && !forcePublic ? '-' : '') + withoutPrefix;
```
Tofandel, line 1277: I thought because of this line, here we return '-' but we should return the original private prefix.

But from testing, it seems there is some magic running after that already and it keeps the original private prefix somehow

lib/src/migrators/module.dart
Tofandel: Yes it's just for the sake of avoiding unecessary changes during a migration, but I definitely don't understand everything that's happening after that, because some tests that look like they should fail with this, don't 🤷 

```scss
<==> arguments
--migrate-deps --remove-prefix=app
<==> input/entrypoint.scss
@import "library";
a {
  color: $app-var;
}

<==> input/_library.scss
$_app-red: red;
$app-var: $_app-red;

<==> output/entrypoint.scss
@use "library";
a {
  color: library.$var;
}

<==> output/_library.scss
$_red: red;
$var: $_red;
```

pamelalozano16 (approved):
LGTM

PR #276, opened 2025-04-09
Add tests and fix behavior of remove prefix

Tofandel:
A few more fixes over #275

Mainly runs unprefixing before unprivatizing a variable to handle cases where the prefix could have leading dashes or underscores


nex3:
It looks like this has some merge conflicts.

PR #280, opened 2025-07-18
Better handling of late imports

### LINKED ISSUE #278: Migrator treats @imports as nested if they are preceded by any rule, regardless of actual nesting
When using the migrator for modules, if there is any rule in the file preceding an @import directive for a partial which uses sass variables declared in that file, the tool fails. The following error message is produced:

> Error: This stylesheet was loaded by a nested import in minimal.sass. The module system only supports loading nested CSS using the load-css() mixin, which doesn't allow access to variables from the outer stylesheet.

The error is not documented, and I couldn't find it anywhere in this repo outside of a test spec which didn't offer much more insight into the cause. In this failure case, it's a little confusing as output, because there are no nested imports.

Expected behaviour: The migrator does not treat non-nested imports as nested.

Minimal reproducable example:

minimal.sass:
```sass
$text-color: #fff

body
    background: #000

// Import partials.
@import "minimal/base"

```

_base.sass:
```sass
body
  color: $text-color

```

command:
```bash
$ sass-migrator module -dv ./minimal.sass
```

jathak:
Fixes #278.

Also fixes a bug where a file that uses and imports the same dependency would be migrated to duplicate `@use` rules.

### DIFF
--- a/lib/src/migrators/module.dart
+++ b/lib/src/migrators/module.dart
@@ -35,6 +35,12 @@ class ModuleMigrator extends Migrator {
   final argParser = ArgParser()
     ..addFlag('built-in-only',
         help: 'Migrates global functions without migrating @import.')
+    ..addMultiOption('safe-at-rule',
+        help: 'CSS at-rules (without the @) used by postprocessing tools '
+            'that should not be considered to be emitting CSS.')
+    ..addFlag('unsafe-hoist',
+        help: 'Allow the migrator to hoist late imports to the top of the '
+            'file even when they emit CSS.')
     ..addMultiOption('remove-prefix',
         abbr: 'p',
         help: 'Removes PREFIX from all migrated member names.\n'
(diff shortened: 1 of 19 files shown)


lib/src/migrators/module.dart
```diff
@@ -816,23 +843,34 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
             visitDependency(ruleUrl, import.span);
             _upstreamStylesheets.remove(currentUrl);
           }
         } else if (_useAllowed) {
-          migratedRules.addAll(_migrateImportToRules(ruleUrl, import.span));
+          inPlaceUseRules.addAll(_migrateImportToRules(ruleUrl, import.span));
+        } else if (!isNested &&
+            (!(references.fileEmitsCss[canonicalUrl] ?? true) ||
+                unsafeHoist &&
+                    references.anyMemberReferenced(canonicalUrl, currentUrl))) {
```
nex3, line 861: Add parens to avoid mixed `||` and `&&`

lib/src/migrators/module.dart
```diff
@@ -894,8 +932,10 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
     if ((rules.isEmpty && normalForwardRules == null) ||
         withClause.isNotEmpty ||
         references.anyMemberReferenced(canonicalUrl, currentUrl)) {
-      _usedUrls.add(canonicalUrl);
-      rules.add('@use $quotedUrl$asClause$withClause');
+      if (!_usedUrls.contains(canonicalUrl)) {
+        _usedUrls.add(canonicalUrl);
```
nex3, line 936:
```suggestion
      if (_usedUrls.add(canonicalUrl)) {
```

nex3 (approved):

PR #282, opened 2025-07-31
Improved late import hoisting

jathak:
While 2.4.0 allowed the migrator to hoist late imports that did not emit CSS themselves, this additionally allows it to hoist them above safe at rules in the same file, even if the imports themselves do actually emit CSS.


CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+## 2.4.2
+
+### Module Migration
+
+* Late `@import` rules can now be hoisted above safe at rules in the same file.
```
nex3, line 5: Nit:

```suggestion
* Late `@import` rules can now be hoisted above safe at-rules in the same file.
```

lib/src/migrators/module.dart
```diff
@@ -188,8 +191,7 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
   /// or `@forward` rule, or null if none has been visited yet.
   FileLocation? _afterLastImport;
 
-  /// Whether @use and @forward are allowed in the current context.
-  var _useAllowed = true;
+  var _useAllowed = UseAllowed.allowed;
```
nex3, line 194: This should still have a doc comment.

lib/src/migrators/module.dart
```diff
@@ -990,7 +998,8 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
   /// `meta.load-css`. The third is a string of variables that should be added
   /// to a `show` clause of a `@forward` rule so that they can be configured by
   /// an upstream file.
-  (Uri, String?, String?) _migrateImportCommon(Uri ruleUrl, FileSpan context) {
+  (Uri, String?, String?) _migrateImportCommon(Uri ruleUrl, FileSpan context,
+      {required bool toLoadCss}) {
```
nex3, line 1002: Document this parameter.

lib/src/migrators/module/use_allowed.dart
```diff
@@ -0,0 +1,44 @@
...
+  static const notAllowed = UseAllowed._('notAllowed');
+
+  /// Identifier for this status
+  final String id;
+
+  const UseAllowed._(this.id);
+
+  /// Returns [requiresHoist] unless [this] is already [notAllowed], in which
+  /// case it should remain [notAllowed].
+  UseAllowed get lowerToRequiresHoist =>
```
nex3, line 36: Since the most natural name for this is a verb phrase, it should be a method rather than a getter.

nex3 (approved):

PR #300, opened 2026-05-20
Support import-only files with include rules

jathak:
This enables migration of some nested/late imports that the migrator couldn't previously handle by first wrapping the file that's depended on via nested import in a mixin and then `@include`ing that mixin in the import-only file.


jathak:
https://github.com/sass/migrator/pull/301 should fix the CI here

test/migrators/module/partial_migration/import_only_include/nested_no_arguments_no_forward.hrx
nex3, line 1: The tests that are exercising nested imports should probably mention that in their titles.

lib/src/migrators/module/references.dart
```diff
@@ -701,6 +709,16 @@ class _ReferenceVisitor extends ScopedAstVisitor {
     } else if (namespace == null) {
       _unresolvedReferences[node] = currentScope;
     }
+    if (isImportOnlyFile(_currentUrl)) {
+      if (_importOnlyIncludes.containsKey(_currentUrl)) {
+        throw MigrationSourceSpanException(
+            "Found a second @include rule in an import-only file. "
+            "Import-only files should contain at most one @include.",
+            node.span);
```
nex3: Would it be possible to support multiple rules if we stored `_importOnlyIncludes`' values as lists?

lib/src/migrators/module.dart
```diff
@@ -990,6 +1013,41 @@ class _ModuleMigrationVisitor extends MigrationVisitor {
...
+    }
+    var namespace = _namespaceForDeclaration(declaration);
+    var args = [
+      for (var arg in rule.arguments.positional)
+        switch (arg) {
+          VariableExpression(:var name, namespace: _?) =>
+            '${_namespaceForDeclaration(references.variables[arg]!)}.\$$name',
+          VariableExpression(:var name) => '\$$name',
+          _ => throw MigrationSourceSpanException(
+              'Import-only @include rules may only pass variables as arguments.',
```
nex3: Why this restriction? It seems like at the very least it wouldn't hurt to allow literal values, and really the only semantically difficult argument I can think of would be user-defined functions (which function more or less like variables anyway).

lib/src/migrators/module/references.dart
jathak: Yeah, I initially avoided it because I thought it would be complicated turning one `@import` rule into multiple `@include` rules, but I forgot we already had logic for turning one rule into multiple because an `@import` rule can contain multiple imports. Done.

lib/src/migrators/module.dart
jathak: I initially restricted it to avoid having to recursively serialize all expressions (making sure to add any additional `@use` rules as necessary), but I ended up writing a serializer that can add/change namespaces as necessary, so that eliminates the need for this restriction.

test/migrators/module/partial_migration/import_only_include/nested_no_arguments_no_forward.hrx
jathak: Done

jathak (commented):
Added `NamespacedSerializer` utility class to allow for arbitrary arguments to be copied from the `@include` rule in the import-only file. It does copy a fair bit of logic from `toString()` methods on some of the AST nodes, so let me know if there's a better way to implement this to avoid the duplicated code.

lib/src/util/namespaced_serializer.dart
nex3, line 1: Consider publicly exposing a function (or several functions) from this file rather than a class, since in practice the class is only ever used to call a single method and then discarded.

lib/src/util/namespaced_serializer.dart
```diff
@@ -0,0 +1,192 @@
...
+//
+// Use of this source code is governed by an MIT-style
+// license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import 'package:charcode/charcode.dart';
+import 'package:sass_api/sass_api.dart';
+
+/// A utility for serializing Sass expressions while adding or changing
+/// namespaces of Sass references.
```
nex3: This means that the migrator now effectively has two different ways of modifying an expression: one by patching, and now this which serializes from scratch. Is it possible to consolidate these back down into a patch? For example, could you represent this as one patch that adds the original text of the `@include` from the import-only file, followed by additional patches that adjust its namespaces (possibly using something like Dart Sass's interpolation maps to realign spans from the original import-only file to the new `@include` location)? My two practical concerns here are:

1. This is duplicating a lot of the Sass inspect/serialization logic, which means more work you'll have to do any time the Sass AST changes.
2. This throws away any formatting/comments/etc from the original source.

If you think this distinction makes more sense than trying to make it all work via patching, I do think it's probably worth including in the documentation some discussion of when this is appropriate relative to the original patch-based approach.

lib/src/util/namespaced_serializer.dart
```diff
@@ -0,0 +1,192 @@
...
+            '${serializeInterpolation(name)}'
+                '(${serializeArgumentList(arguments)})',
+          LegacyIfExpression(:var arguments) =>
+            'if(${serializeArgumentList(arguments)})',
+          ListExpression(:var contents, :var hasBrackets, :var separator) =>
+            (hasBrackets ? '[' : '') +
+                [for (var item in contents) serialize(item)].join(
+                  switch (separator) {
+                    .comma => ', ',
+                    .space => ' ',
```
nex3:
```suggestion
                    .space || .undecided => ' ',
```

test/migrators/module/partial_migration/import_only_include/nested_multiple_includes.hrx
nex3, line 1: As before, since this is testing nested includes it should mention that in its title.

nex3 (changes requested):

lib/src/util/namespaced_serializer.dart
jathak: As is, the migrator won't allow patches like this, since the patch offsets are always based on the original file, so after patching in the original text, it's not possible to add additional patches within it, but I do agree that the patch-based approach makes more sense if at all possible, so I'll look into adding something that can do the inverse: first patch the span of the original `@include` (without actually patching the underlying source file) and then apply that patch that patched span into the import-only file.

lib/src/util/namespaced_serializer.dart
jathak: Done

test/migrators/module/partial_migration/import_only_include/nested_multiple_includes.hrx
jathak: Done

jathak (commented):
Updated the logic here to use patching instead of serialization.

nex3 (approved):

PR #305, opened 2026-07-08
Create a branch that uses Dart Sass 2.0.0

### LINKED ISSUE #296: Gracefully handle obsolete migrations
The migrator is in a bit of a sticky place when it comes to migrations for (parse-time) behavior that has already changed in Dart Sass 2. Because the migrator uses Dart Sass as its parser, if Dart Sass starts to emit parse errors for a given deprecated syntax, new versions of the migrator won't be able to migrate it.

At the very least, we should mitigate this by having older migrations print messages directing users to the last version of the migrator that supported that migration. Ideally, we would do something even more clever like installing and running those older versions for a period of time after the breaking change lands (or just shipping them with the migrator). It's also possible that we could do some renaming and have multiple versions of the Sass parser coexist within the same process to continue supporting older migrations for a time.

We'll need to decide which approach to take an implement it before Dart Sass 2 releases.

nex3:
Closes #296


lib/src/exception.dart
```diff
@@ -4,8 +4,8 @@
 // license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
-import 'package:sass_api/sass_api.dart';
 import 'package:source_span/source_span.dart';
+import 'package:sass_api/sass_api.dart';
```
jathak, line 7: Imports should be sorted alphabetically

jathak (approved):

PR #306, opened 2026-08-20
Update Dart SDK constraint to 3.13.0

nex3:


bin/sass_migrator.dart
```diff
@@ -11,8 +11,8 @@ import 'package:sass_migrator/src/runner.dart';
 
 // We can't declare args as a List<String> or Iterable<String> beacause of
 // dart-lang/sdk#36627.
```
jathak: Is this comment still true?

lib/src/migrators/module/reference_source.dart
```diff
@@ -73,39 +77,44 @@ final class UseSource extends ReferenceSource {
...
 /// built-in module.
-final class BuiltInSource extends ReferenceSource {
-  final Uri url;
+final class BuiltInSource(String module) extends ReferenceSource {
+  @override
+  final Uri url = Uri.parse("sass:$module");
 
   /// Constructs a [BuiltInSource] for a [module].
-  BuiltInSource(String module) : url = Uri.parse("sass:$module");
+  this;
```
jathak: This doc comment doesn't really say anything non-obvious, so it's probably best to just eliminate it

jathak (approved):

bin/sass_migrator.dart
nex3: Looks like it's not.

